PR #10, opened 2016-04-20
Handle transactions better

ariejan:
**Current situation**

A transaction is added, the balance is updated and the next goal is marked as achieved if necessary.

**New situation**

A transaction is added, the balance is recalculated given all transactions for that balance, all goals are achieved that need achieving/unachieving.

- [x] Refactor calculating balance and goals
- [ ] Trigger `KudoKalkulator` on goal/transaction mutations.


ariejan:
Never mind all this.

PR #36, opened 2016-12-16
Kudo on focus

CanolKabisa:


app/models/user.rb
```diff
@@ -11,6 +11,7 @@ class User < ActiveRecord::Base
 
   def self.from_omniauth(access_token)
     data = access_token.info
+    return if data['email'].split('@')[1] != 'kabisa.nl'
```
thnukid: `unless` is misschien duidelijker hier, eg:
`return unless data['email'].split('@')[1] == 'kabisa.nl' `

Basically moving the focus from the `!=` to `unless`. `==` expresses more what you want to say here.

PR #54, opened 2017-01-04
WIP: Responsive webdesign updates

CanolKabisa:
Added gemfile Will paginate, dashboard, feed, form & guidelines refactor
Added the amount of kudos to the dynamic guidelines div

### DIFF
--- a/app/models/transaction.rb
+++ b/app/models/transaction.rb
@@ -40,9 +40,7 @@ class Transaction < ActiveRecord::Base
   def self.guidelines_between(from, to)
     gl = []
     GUIDELINES.each do |g|
-      gl.push g[0] if g[1] >=from and g[1] <=to
-      gl.push g[1] if g[1] >=from and g[1] <=to
-
+      gl.push g if g[1] >=from and g[1] <=to
     end
     gl
   end
--- a/app/views/dashboard/index.html.haml
+++ b/app/views/dashboard/index.html.haml
@@ -56,14 +56,11 @@
 
         = simple_form_for @transaction do |f|
           .span
-            = t('dashboard.i_give');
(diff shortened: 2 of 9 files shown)


app/models/transaction.rb
```diff
@@ -40,9 +40,7 @@ class Transaction < ActiveRecord::Base
   def self.guidelines_between(from, to)
     gl = []
     GUIDELINES.each do |g|
-      gl.push g[0] if g[1] >=from and g[1] <=to
-      gl.push g[1] if g[1] >=from and g[1] <=to
-
+      gl.push g if g[1] >=from and g[1] <=to
```
seamyd: I think && (over and) is more appropriate here, see: http://stackoverflow.com/questions/1426826/difference-between-and-and-in-ruby

app/views/dashboard/index.html.haml
```diff
@@ -56,14 +56,11 @@
 
         = simple_form_for @transaction do |f|
           .span
-            = t('dashboard.i_give');
-            = f.input_field :receiver,  as: :autocomplete, source_query: autocomplete_user_name_users_url, field_id: false,  placeholder: 'receiver',required: true , :autofocus => true
+            = f.input_field :receiver,  as: :autocomplete, source_query: autocomplete_user_name_users_url, field_id: false,  placeholder: 'To',required: true , :autofocus => true
```
seamyd: make your code more readable by always using the same rules:
* in this case if you provide a hash as argument do: <key><comma><space><value> instead of <key><comma><value> or <key><space><comma><space><value> this is called style (https://github.com/airbnb/ruby) and your eyes will thank you for it ;)

PR #83, opened 2017-03-02
Tweak feed output and simplify implementation

pascalw:
* Don't care about showing newest first in feed, makes stuff easier.
* Set fixed feed title, don't include root_url which looks ugly
* Override updated timestamps so updates don't reappear in the feed


pascalw:
@ariejan sorry went a bit back and forth :D I'm going to leave it at this now.

PR #84, opened 2017-03-06
Added slack webhook

egonm12:


app/models/transaction.rb
```diff
@@ -58,4 +62,46 @@ def receiver_image
     receiver_id.nil? ? '/kabisa_lizard.png' : receiver.picture_url
   end
 
+  def send_slack_notification
+    notifier = Slack::Notifier.new ENV['WEBHOOK_URL']
+
+    notifier.ping(
+        :channel => '#kudo',
+        :attachments => [
```
thnukid: Have a look at `rubocop gem` https://github.com/bbatsov/rubocop

In this hash you can use the new ruby hash syntax

```ruby
{ key: 'value, 'nested_key': { another_key: 'value' } }
```

rubocop will help you

Gemfile
```diff
@@ -1,12 +1,14 @@
...
 
-ruby '2.3.1'
+ruby '2.4.0'
 
 # Rails
 gem 'rails', '~> 5.0.0', '>= 5.0.0.1'
 
 gem 'dotenv-rails', group: [:development, :test]
 
+gem 'slack-notifier'
```
pascalw: It's best practise to specify a (semi) specific version of gems. This to prevent that the gem get's updated to a new major version when you run `bundle update`.

app/models/goal.rb
```diff
@@ -17,6 +17,7 @@ def achieved?
 
   def achieve!
     touch(:achieved_on)
+
```
pascalw: Looks like you introduced some unnecessary whitespace here.

app/models/transaction.rb
```diff
@@ -58,4 +62,46 @@ def receiver_image
     receiver_id.nil? ? '/kabisa_lizard.png' : receiver.picture_url
   end
 
+  def send_slack_notification
+    notifier = Slack::Notifier.new ENV['WEBHOOK_URL']
```
pascalw: You generally want to use `ENV.fetch` to retrieve environment variables. If the env var is not set `ENV['WEBHOOK_URL`] will simply return nil whereas `ENV.fetch` will throw if the var is not set.

Also you'd want to make sure that during application boot the ENV var is fetched, to prevent the application can even boot with a missing environment variable.

app/models/transaction.rb
```diff
@@ -58,4 +62,46 @@ def receiver_image
...
+
+    notifier.ping(
+        :channel => '#kudo',
+        :attachments => [
+            {
+                :fallback => 'New transaction',
+                :color => '#B58342',
+                :pretext => '<!channel> A new kudo transaction has been made! <https://kudos.kabisa.io/|Click here> for more details.',
```
pascalw: Instead of hardcoding the url here you could use the `root_url` helper. This way it will work in any environment you might run this application.

app/models/transaction.rb
pascalw: Also I think you'd want to disable slack notifications in development.

app/models/transaction.rb
```diff
@@ -58,4 +62,46 @@ def receiver_image
...
+                        :short => true
+                    },
+                    {
+                        :title => 'Kudos given for',
+                        :value => self.activity.name.capitalize,
+                        :short => true
+                    },
+                    {
+                        :title => 'Amount of Kudos',
+                        :value => '₭ ' + self.amount.to_s,
```
pascalw: There's a `number_to_kudos` helper that will format the amount for you.

app/models/transaction.rb
```diff
@@ -58,4 +62,46 @@ def receiver_image
...
+                        :short => true
+                    }
+                ],
+                :footer => 'Kabisa | ₭udos Platform | ' + self.created_at.to_s,
+                :footer_icon => 'https://pbs.twimg.com/profile_images/2203769368/kabisa_lizard_400x400.png',
```
pascalw: To keep the application generic I think you should make this configurable. Same for `Kabisa` in the footer.

app/models/user.rb
```diff
@@ -26,6 +26,7 @@ def self.from_omniauth(access_token)
 
   def to_s
     name
+    slack_name
```
pascalw: What's the intention of this? You're now hiding the `name` output, since only the last expression will be returned

env.example
```diff
@@ -1,2 +0,0 @@
-GOOGLE_CLIENT_ID=""
```
pascalw: Any reason to remove this file? IMO it's best practise to clearly document required environment variables.
On that note please add the slack webhook variable here as well.

app/models/transaction.rb
pascalw: Also please make the env var more descriptive, so maybe name it `SLACK_WEBHOOK_URL` or something like that.

app/models/user.rb
thnukid: Since ruby returns the last statement as a default, you will get `slack_name`. 

So `name` can be deleted if that is your intention (dunno if `name` is used somewhere else specific and you now breaking the app).

If you want both you can do:

```ruby
[name, slack_name].join(' ')
```

You might also want to add a test for this ;)

app/models/transaction.rb
```diff
@@ -58,4 +62,46 @@ def receiver_image
     receiver_id.nil? ? '/kabisa_lizard.png' : receiver.picture_url
   end
 
+  def send_slack_notification
```
pascalw: It would be preferable to extract this behaviour into a separate class that handles sending slack notifications, instead of putting this in the model. Functionality in the model should be restricted to things that actually concern the (domain) model.

While you're at it would also be nice to make the notification asynchronous for the following reasons:
* Users shouldn't have to wait until a notification is sent
* If the notification fails to send you don't want the transaction to rollback
* In a background job you'd have to possibility to retry sending the notification

pascalw: Last I would to challenge you to write a (integration) test for this.

app/models/transaction.rb
pascalw: Also disable notifications in tests please

pascalw:
Build is failing @egonm12, see https://travis-ci.org/kabisa/kudo-o-matic/builds/208885384

pascalw:
@ariejan looks like travis is running twice? Why is that?

env.example
ariejan: Second! Do not remove these fields in the `env.example` as it shows others what env variables need to be setup.

app/models/user.rb
egonm12: @pascalw Forgot to delete this, doesn't have any functionality.

app/models/transaction.rb
egonm12: @pascalw When writing `value: '₭ ' + number_to_kudos(transaction.amount),` it says `undefined method 'number_to_kudos' for SlackNotifications:Class`

app/models/transaction.rb
egonm12: @pascalw @pascalw When writing `pretext: "<!channel> A new kudo transaction has been made! <#{root_url}|Click here> for more details.",` it says `undefined local variable or method 'root_url' for SlackNotifications:Class` 

So instead of doing that I was thinking about `pretext: "<!channel> A new kudo transaction has been made! <#{ENV['ROOT_URL']}|Click here> for more details.",`

app/models/transaction.rb
pascalw: Helper methods are only available in views and controllers.
There are ways to include helpers into other classes, but my suggestion would be to make this a static utility method so it can be used easily in any context.

app/models/transaction.rb
pascalw: Same as my previous comment, tho in this case you can't define the helper yourself so your only option would be to include the routing helpers into your service:

```
include Rails.application.routes.url_helpers
```

Gemfile
```diff
@@ -1,12 +1,16 @@
...
+ruby '2.4.0'
 
 # Rails
 gem 'rails', '~> 5.0.0', '>= 5.0.0.1'
 
 gem 'dotenv-rails', group: [:development, :test]
 
+gem 'slack-notifier', '~> 2.1.0'
+
+gem 'rubocop', '~> 0.47.1', require: false
```
pascalw: In the future try to keep pull requests focused and as small as possible (to aid the review process). Rubocop has nothing to do with slack notifications so that should go into a separate PR. Don't worry about it now just keep it in mind next time :-)

app/models/transaction.rb
pascalw: I see it's still synchronous. Are you planning to pick this up in this PR or are you going to tackle this separately? In any case you could probably use http://guides.rubyonrails.org/active_job_basics.html with the default async threaded backend. That's good enough for this since you can afford to lose a notification if there happens to be a job in the queue during deployment.

app/services/slack_notifications.rb
```diff
@@ -0,0 +1,44 @@
+require 'slack-notifier'
+
+class SlackNotifications
+
+  def self.send_kudo_transaction(transaction)
+    return unless Rails.env == 'development'
```
pascalw: I don't think this is what you want? It will now only send notifications in development.

You could instead only send when the environment is not development or test, or only send when the `SLACK_WEBHOOK_URL` env var is set.

app/services/slack_notifications.rb
```diff
@@ -0,0 +1,44 @@
...
+    notifier = Slack::Notifier.new ENV.fetch('SLACK_WEBHOOK_URL')
+
+    notifier.ping(
+        channel: '#testkudo',
```
pascalw: Would be nice to make this configurable via ENV as well

app/services/slack_notifications.rb
```diff
@@ -0,0 +1,44 @@
...
+                        short: true
+                    },
+                    {
+                        title: 'Kudos given for',
+                        value: transaction.activity_name.capitalize,
+                        short: true
+                    },
+                    {
+                        title: 'Amount of Kudos',
+                        value: '₭ ' + transaction.amount.to_s,
```
pascalw: Please try using the helper to format kudos. Formatting kudos for display is a core concept that you don't want to scatter all of the code base.

pascalw:
Hey @egonm12 can you update the travis CI build config so the test can run again? It's now failing with this:

```
Your Ruby version is 2.3.1, but your Gemfile specified 2.4.0
```

PR #88, opened 2017-03-09
Feature db slack name

egonm12:


app/assets/stylesheets/dashboard.scss
```diff
@@ -62,8 +75,8 @@ li {
 
 .new-visitor-show{
   width: 100vw;
-  height: 100vh;
-  position: fixed;
+  height: 170vh;
```
pascalw: Why `170vh`? `100vh` makes sense but `170vh` seems arbitrary?

app/controllers/transactions_controller.rb
```diff
@@ -5,8 +5,16 @@ def new
   end
 
   def create
-    @transaction = TransactionAdder.create(params[:transaction], current_user)
-    flash[:error] = @transaction.errors.full_messages.to_sentence
+    if params[:transaction][:activity].empty? && params[:transaction][:receiver].empty?
```
pascalw: Usually validations like these are implemented in the model. Reason is that the kinds of validations usually apply to the model, regardless of the context where it's used. By putting the validations here in the controller you can't reuse them elsewhere. See http://guides.rubyonrails.org/active_record_validations.html for details.

app/services/slack_notifications.rb
```diff
@@ -0,0 +1,75 @@
...
+
+   attr_reader :transaction
+
+   def initialize(transaction)
+     @transaction = transaction
+   end
+
+   def notify_slack!
+     send_to_channel
+     send_to_user
```
pascalw: Maybe instead of sending two notifications you could `@mention` people on the public channel? Since the public message already contains the names of people it's likely they'll receive a notification on the channel anyway, so with this those people would receive 2 notifications for the same event.

config/application.yml
```diff
@@ -0,0 +1,5 @@
+company:
+  footer_name: Kabisa
```
pascalw: These settings seem to apply to slack notifications only. It would be clearer to reflect this in the setting names?

db/schema.rb
```diff
@@ -87,6 +87,7 @@
     t.string   "provider"
     t.string   "uid"
     t.string   "avatar_url"
+    t.string   "slack_name"
```
pascalw: The migration for this field seems to missing?

app/models/transaction.rb
```diff
@@ -58,4 +61,11 @@ def receiver_image
     receiver_id.nil? ? '/kabisa_lizard.png' : receiver.picture_url
   end
 
+  def send_slack_notification
+    SlackNotifications.new(self).notify_slack!
+  end
+
+  def subtract_from_balance
```
pascalw: I'm missing a unit test here. Having the balance be correct is pretty important so please make sure this is properly covered by (unit) tests.

app/assets/stylesheets/dashboard.scss
egonm12: @pascalw Temporary fix to cover the whole text field and changed the position to 'absolute' so you can close the pop-up when using mobile.

app/controllers/transactions_controller.rb
egonm12: @pascalw Any suggestions for doing this? Because the original code was in the transaction model `validates :activity, presence: true` but that doesn't work because the transaction always creates an activity_id, even when the activity is empty.

app/assets/stylesheets/dashboard.scss
pascalw: Ok what definitive fix are you planning on, then?

app/controllers/transactions_controller.rb
pascalw: The problem then seems to be that activity allows to be created without a name.

db/schema.rb
pascalw: But you didn't push that commit?

egonm12: @pascalw Everything is now pushed in pull request 84. Andy merged this pull with 84.

app/assets/stylesheets/dashboard.scss
egonm12: Maybe something like this http://getbootstrap.com/javascript/#modals

PR #138, opened 2017-06-08
Transaction filter function implemented

egonm12:


spec/features/filter_transaction_spec.rb
```diff
@@ -0,0 +1,64 @@
+require 'rails_helper'
+require 'capybara/poltergeist'
+Capybara.javascript_driver = :poltergeist
+
+options = {js_errors: false, timeout: 30}
+Capybara.register_driver :poltergeist do |app|
+  Capybara::Poltergeist::Driver.new(app, options)
+end
+
+module WaitForAjax
```
thnukid: I am guessing that you want to test for ajax request, and "somehow" the ajax request takes to long.

https://nulogy.com/who-we-are/company-blog/articles/sleep-is-for-the-weak/

TLDR; Capybara will wait, when u using the right matchers, like http://www.rubydoc.info/github/jnicklas/capybara/Capybara/RSpecMatchers

egonm12: @thnukid gonna look into this with Remco next week. Don't have any experience with testing ajax requests in a decent way. Thanks for the feedback anyway 👍 , gonna check this out 😄 !

Issue #124, opened 2017-05-31
Emoji's in autocomplete are hosted on external domain


pascalw:
Emoji's in autocomplete are hosted on on yuku-t.com.
This is kind of shady :-)

pascalw:
@egonm12 is this stil on your radar? IMO this should have high priority.

egonm12:
@pascalw yes, in the backlog for this sprint

PR #154, opened 2017-06-20
Host emoji's for fixing issue #124

egonm12:


env.example
```diff
@@ -6,3 +6,4 @@ ROOT_URL="https://homepage.org"
 COMPANY_USER="COMPANY_USERNAME"
 RAILS_ENV="production"
 RACK_ENV="production"
+S3_BUCKET_EMOJI_URL="https://mybucket.s3.amazonaws.com/emojis/"
```
pascalw: I think this is no longer needed, right?

PR #166, opened 2017-06-29
Changed seed file for standard template and added rake task to promot…

egonm12:
…e first user to admin


egonm12:
A plan to improve this feature is the option to log in without a Google account and create a standard admin user with the seed file.

PR #169, opened 2017-06-29
Updatet Slack Reminder text

egonm12:


egonm12:
@remcohh Is it ok to set the secret key in `devise.rb`?

PR #191, opened 2017-09-19
Added checks to prevent the last administrator from being removed

Laugslander:
Implemented (and integration tested) checks that prevent the last administrator from being removed from the system.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,7 +1,4 @@
 class User < ActiveRecord::Base
-  after_update :ensure_an_admin_remains
-  after_destroy :ensure_an_admin_remains
-
   acts_as_voter
 
   has_many :sent_transactions, class_name: 'Transaction', foreign_key: :sender_id
@@ -36,13 +33,5 @@ class User < ActiveRecord::Base
 
   def picture_url
     avatar_url || '/no-picture-icon.jpg'
-  end
-
-  private
-
-  def ensure_an_admin_remains
-    if User.where(admin: true).count < 1
-      raise "Last administrator can't be removed from the system"
-    end
   end
 end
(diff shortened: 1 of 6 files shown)


spec/features/modal_spec.rb
```diff
@@ -2,7 +2,7 @@
 require 'capybara/poltergeist'
 Capybara.javascript_driver = :poltergeist
 
-options = {js_errors: false, timeout: 30}
+options = {js_errors: false, timeout: 2.minutes}
```
matthijsgroen: PhantomJS (and thus poltergeist) is EOL. https://github.com/ariya/phantomjs/issues/15105
I would suggest to make a future PR to switch to another driver, like chrome headless.

app/models/user.rb
```diff
@@ -1,33 +1,32 @@
 class User < ActiveRecord::Base
-acts_as_voter
+  after_update :ensure_an_admin_remains
+  after_destroy :ensure_an_admin_remains
```
matthijsgroen: Wouldn't it be cleaner to 'abort' `before_update`/`before_destroy` ? Isn't there a way to just prevent the action (by refusing to set admin to false, or removing the user) and not use exceptions as control flow?

This means extra logic should be in the controller though to detect the action is aborted and show the correct flash message / redirect. But using exceptions for logic control flow instead of 'errors' feels weird.

spec/features/modal_spec.rb
Laugslander: Created an issue for replacing Poltergeist in the near future.

app/models/user.rb
Laugslander: Moved the admin removal check logic to the Admin::UsersController to prevent the exception control flow.

PR #192, opened 2017-09-26
Added a REST API with authentication for retrieving balances and goals

Laugslander:
Implemented a REST API using the JSON API standard for retrieving Kudo balances and goals. The API is protected with simple token based authentication.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -54,6 +54,10 @@ gem 'redcarpet', '~> 3.3', '>= 3.3.4'
 
 gem 'timecop', '~> 0.8.1'
 
+gem 'jsonapi-resources'
+gem 'jsonapi-resources-matchers'
+gem 'google-id-token'
+
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'byebug'
--- a/Gemfile
+++ b/Gemfile
@@ -54,10 +54,10 @@ gem 'redcarpet', '~> 3.3', '>= 3.3.4'
 
 gem 'timecop', '~> 0.8.1'
 
-gem 'jsonapi-resources'
-gem 'jsonapi-resources-matchers'
-gem 'google-id-token'
-gem 'rack-cors'
+gem 'jsonapi-resources', '~> 0.9'
+gem 'jsonapi-resources-matchers', '~> 1.0'
+gem 'google-id-token', '~> 1.4'
+gem 'rack-cors', '~> 1.0'
 
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
--- a/Gemfile
+++ b/Gemfile
@@ -57,6 +57,7 @@ gem 'timecop', '~> 0.8.1'
 gem 'jsonapi-resources'
 gem 'jsonapi-resources-matchers'
 gem 'google-id-token'
+gem 'rack-cors'
 
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
--- a/app/controllers/api/v1/authentication_controller.rb
+++ b/app/controllers/api/v1/authentication_controller.rb
@@ -1,0 +1,32 @@
+class Api::V1::AuthenticationController < ActionController::Base
+  # before_action :add_allow_credentials_headers
+
+  respond_to :json
+
+  def retrieve_api_token
+    if validate_jwt_token(params[:jwt_token])
+      user = User.from_api_token_request(api_token_request_params)
+      render json: {api_token: user.api_token}
+    else
+      render json: {error: 'Invalid ID token'}
+    end
+  end
+
+  private
+
+  def validate_jwt_token(jwt_token)
+    validator = GoogleIDToken::Validator.new
+
+    begin
+      validator.check(jwt_token, ENV["KUDO_O_MOBILE_CLIENT_ID"])
+      true
+    rescue GoogleIDToken::SignatureError => e
+      logger.error e.message
+      false
+    end
+  end
+
+  def api_token_request_params
+    params.permit(:jwt_token, :uid, :provider, :name, :email, :avatar_url)
+  end
+end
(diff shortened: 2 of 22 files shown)


app/controllers/api/v1/authentication_controller.rb
```diff
@@ -0,0 +1,32 @@
+class Api::V1::AuthenticationController < ActionController::Base
+  # before_action :add_allow_credentials_headers
+
+  respond_to :json
+
+  def retrieve_api_token
+    if validate_jwt_token(params[:jwt_token])
```
pascalw: Who's going to call this API? It requires a JWT token - who is providing this token?

app/models/user.rb
```diff
@@ -1,33 +1,37 @@
...
+        avatar_url: data['image']
     )
 
+    UserMailer.new_user(user)
+
+    user
+  end
+
+  def self.from_api_token_request(params)
+    user = User.where(uid: params['uid']).first_or_create(params.except(:jwt_token))
```
pascalw: Does it make sense to create a user during an API token request? Sounds fishy 🤔

spec/requests/api/v1/balances_controller_spec.rb
```diff
@@ -0,0 +1,215 @@
...
+    let! (:balance1) {create(:balance)}
+    let! (:balance2) {create(:balance)}
+
+    context 'with valid api-token' do
+      let (:user) {create(:user, api_token: 'X0EfAbSlaeQkXm6gFmNtKA')}
+
+      it 'returns all balances' do
+        get request, headers: {'Api-Token': user.api_token}
+
+        expect(json['data'][0]['id']).to eq(balance1.id.to_s)
```
pascalw: This is OK, but since this API is going to be a contract (between the Kudo server and some client(s)) I would rather have the expectations more exact, I.E. matching the entire response. This to prevent accidental API breakage in the future, for example by removing a field not included in the expectation currently.

spec/requests/api/v1/balances_controller_spec.rb
```diff
@@ -0,0 +1,215 @@
...
+        expect(json['data']['attributes']['name']).to eq(balance.name)
+        expect(json['data']['attributes']['current']).to eq(balance.current)
+      end
+
+      it 'persists the newly created balance' do
+        expect {
+          post request,
+               headers: {'Api-Token': user.api_token, 'Content-Type': 'application/vnd.api+json'},
+               params: {data: {type: 'balances', attributes: {name: balance.name, current: balance.current}}}.to_json
+        }.to change {Balance.count}
```
pascalw: Also here I think you could be a bit more precise in your expectation, to verify that all JSON fields are mapped properly. (So include all fields in the expectation).

spec/requests/api/v1/balances_controller_spec.rb
```diff
@@ -0,0 +1,215 @@
...
+    context 'with valid api-token' do
+      let (:user) {create(:user, api_token: 'X0EfAbSlaeQkXm6gFmNtKA')}
+
+      it 'returns the updated balance associated with the id' do
+        patch request,
+              headers: {'Api-Token': user.api_token, 'Content-Type': 'application/vnd.api+json'},
+              params: {data: {type: 'balances', id: balance.id, attributes: {name: 'edited name', current: true}}}.to_json
+
+        expect(json['data']['attributes']['name']).to eq('edited name')
+        expect(json['data']['attributes']['current']).to eq(true)
```
pascalw: Also here it would be good to test that the PATCH request updates only the fields that are passed in the request body and leaves any field _not_ included untouched.

spec/support/request_helpers.rb
```diff
@@ -0,0 +1,11 @@
+module Requests
+  module JsonHelpers
+    def json
+      JSON.parse(response.body).with_indifferent_access
+    end
+
+    def expect_unauthorized
```
pascalw: Since the API is a contract, please include the HTTP status code in your expectation as well. This is very important for REST clients.

app/controllers/api/v1/goals_controller.rb
```diff
@@ -0,0 +1,9 @@
+class Api::V1::GoalsController < Api::V1::ApiController
+  def next_amount
```
pascalw: Is it useful to have the next amount and name separate? My guess would be that API clients would need both the name and amount at the same time, so you might want to combine this into one API call.

Also the path of this resources perhaps could be `GET /api/v1/goals/next`. Where `next` is a symbolic primary key that identifies the next goal.

spec/requests/api/v1/goals_controller_spec.rb
```diff
@@ -0,0 +1,250 @@
...
+    end
+  end
+
+  describe 'GET api/v1/goals/next_amount' do
+    let (:request) {'/api/v1/goals/next_amount'}
+    let! (:balance) {create(:balance)}
+    let! (:goal1) {create(:goal)}
+    let! (:goal2) {create(:goal)}
+
+    context 'with valid api-token' do
```
pascalw: What would happen when there is no next goal? Might want to add a test for this case too.

Gemfile
```diff
@@ -54,6 +54,11 @@ gem 'redcarpet', '~> 3.3', '>= 3.3.4'
 
 gem 'timecop', '~> 0.8.1'
 
+gem 'jsonapi-resources'
```
pascalw: It's generally preferable to specify versions of your dependencies explicitly. If you don't, then gems will likely get unintentionally updated at some point, potentially breaking your app.

It's fine to leave a little fuzzy about versions, for example by specifying `'gem 'jsonapi-resources', '~> 0.9'. This would allow 'safe' updates in the `0.9.x` range, but disallow larger updates like `0.10` or `1.0`.

app/controllers/api/v1/authentication_controller.rb
Laugslander: The Kudo-o-Mobile app will call this function. The app provides a JWT token that it receives from signing in using Google OAuth2. The backend validates this token.

app/models/user.rb
Laugslander: Te API token is stored in the User table in database. When a new user signs into the the mobile app (before signing into the web app), he doesn't have a User record yet.

Laugslander:
- Improved and refactored the naming of the balance- and goal specific non-CRUD API controller actions. 

- Improved and refactored the integration tests by making the response tests more specific and adding tests for HTTP status codes.

- Added versions to the API related gems in the Gemfile file (also created an issue for this).

PR #308, opened 2018-04-13
Sign-up and sign-in

roedesh:
The following elements were implemented:

- Registration form with email validation
- Sign-in form
- Added tests for both registration and sign-in forms
- Modified transaction tests because there no longer is a Google login
- Added doorkeeper so Jeroen can authenticate via mobile


app/controllers/registrations_controller.rb
```diff
@@ -0,0 +1,8 @@
+class RegistrationsController < Devise::RegistrationsController
+
+  private
+
+  def after_inactive_sign_up_path_for(_resource)
```
ariejan: This forces a user to sign in after signing up. Devise has an option (check the config) to allow a user to sign in with an unconfirmed account for a period of time. 

How about we sign a user in directly after signing up and and allow them 24 hours to confirm their email address?

app/models/user.rb
```diff
@@ -6,7 +6,7 @@ class User < ActiveRecord::Base
   # :confirmable, :lockable, :timeoutable and :omniauthable
   # :database_authenticatable, :registerable,
   # :recoverable, :rememberable, :trackable, :validatable
-  devise :omniauthable, :omniauth_providers => [:google_oauth2, :slack]
+  devise :omniauthable, :registerable, :confirmable, :database_authenticatable, :validatable, :omniauth_providers => [:google_oauth2, :slack]
```
ariejan: Is the `google_oauth2` oauth provider still relevant? I can understand Slack, to link your account, but google should not be necessary any longer.

app/views/devise/shared/_links.html.erb
```diff
@@ -0,0 +1,25 @@
...
+
+<%- if devise_mapping.confirmable? && controller_name != 'confirmations' %>
+  <%= link_to "Didn't receive confirmation instructions?", new_confirmation_path(resource_name) %><br />
+<% end -%>
+
+<%- if devise_mapping.lockable? && resource_class.unlock_strategy_enabled?(:email) && controller_name != 'unlocks' %>
+  <%= link_to "Didn't receive unlock instructions?", new_unlock_path(resource_name) %><br />
+<% end -%>
+
+<%- if devise_mapping.omniauthable? %>
```
ariejan: I think we're not providing any oauth signups/signins any more.

app/views/devise/shared/_links.html.erb
```diff
@@ -0,0 +1,25 @@
...
+
+<%- if devise_mapping.recoverable? && controller_name != 'passwords' && controller_name != 'registrations' %>
+  <%= link_to "Forgot your password?", new_password_path(resource_name) %><br />
+<% end -%>
+
+<%- if devise_mapping.confirmable? && controller_name != 'confirmations' %>
+  <%= link_to "Didn't receive confirmation instructions?", new_confirmation_path(resource_name) %><br />
+<% end -%>
+
+<%- if devise_mapping.lockable? && resource_class.unlock_strategy_enabled?(:email) && controller_name != 'unlocks' %>
```
ariejan: How do we use account locking?

config/initializers/doorkeeper.rb
```diff
@@ -0,0 +1,153 @@
+Doorkeeper.configure do
```
ariejan: This looks like a default configuration files with the bare minimum changes to make it work. Did we consider all options and security measures listed here? For instances, "restrict access to the web interface for adding oauth authorized applications" seems something we'd need to add?

spec/features/add_transaction_spec.rb
```diff
@@ -1,27 +1,44 @@
...
+  }
+  let!(:transaction_2) {
+    Transaction.create sender: user, receiver: user, activity: activity, amount: 10,
+                       balance: balance
+  }
 
   before do
     visit '/sign_in'
-    click_link 'Log in with Google'
+    # fill_in 'transaction_receiver_name', with: 'Harry'
```
ariejan: There's commented out code here. Is it not necessary any longer? Just remove it!

app/views/devise/registrations/new.html.haml
```diff
@@ -31,5 +31,6 @@
               characters minimum
       .form-actions
         = f.submit "Sign up"
-
--# TODO: Add GRPR compliant checkboxes
\ No newline at end of file
+      -# TODO: Add GRPR compliant checkboxes
```
ariejan: Just saw this comment pop up. Could you add it to the backlog? GRPR compliance seems quite relevant to have.

roedesh:
All requested changes have been implemented :-)

PR #311, opened 2018-04-20
Feature 156874214 Email confirmation status

roedesh:
- Added helper function `unconfirmed_access_hours_left` to devise_helper, which returns the amount of hours left for unconfirmed access for a given user.
- Added email confirmation status message `Please confirm your e-mail within <unconfirmed_access_hours_left> hours to keep access to your account. Unconfirmed accounts get deleted after 3 days. Didn't receive confirmation instructions?` to the root page. The last sentence is a link to `new_user_confirmation_path`.
- Added RSpec test for `unconfirmed_access_hours_left`.
- Added styling for a orange variation of the message box, called `warning-message`.
- Added a underline to anchors inside a message box.


app/views/transactions/_message.html.haml
```diff
@@ -1,3 +1,13 @@
+- if current_user.confirmed_at == nil
```
ariejan: I think there is some helper like `current_user.confirmed?` that's easier on the eyes :-). At the very least use the more idiomatic `current_user.confirmed_at.nil?`.

app/views/transactions/_message.html.haml
```diff
@@ -1,3 +1,13 @@
...
+  .message-container.show-message
+    .warning-message.margin-bottom
+      %span.message
+        .fa.fa-exclamation-circle
+        %div
+          Please confirm your e-mail within #{unconfirmed_access_hours_left(current_user)} hours to keep access to your account.
```
ariejan: This reminds me, we need to add a story to the backlog to add I18n, so we can ship this in at least Dutch as well. 

I've added this story already.

app/views/transactions/_message.html.haml
```diff
@@ -1,3 +1,13 @@
...
+          Unconfirmed accounts get deleted after 3 days.
```
ariejan: Could we use Devise's settings here instead of a hard-coded "3 days"? Maybe `time_ago_in_words` or one of its sibling helpers can help?

PR #313, opened 2018-04-25
Feature 156797203 GDPR View Data

roedesh:
- Added a page for viewing personal information at `/account/view_data`, also added buttons to the transaction and votes data page.
- Added a page for viewing `transactions`, with pagination, at `/account/view_data/transactions`
- Added a page for viewing `votes`, with pagination, at `/account/view_data/votes`
- Added a "View my data" button on the account edit page, right beneath the edit form.
- Moved account edit route from `/edit_account` to `/account`

screen4

screen shot 2018-04-25 at 10 23 59

screen2

screen3

### DIFF
--- a/app/assets/stylesheets/components/_forms.scss
+++ b/app/assets/stylesheets/components/_forms.scss
@@ -47,6 +47,7 @@
 }
 
 .actions {
+  margin-bottom: 15px;
   .btn {
     background-color: $base-color-brand;
     color: #eee;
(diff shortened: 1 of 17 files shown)


ariejan:
Looks good code-wise. Same comment about the `Transaction.for_user(@user)` instead of the more readable `current_user.transactions`. 

Also, I'm not sure I understand the columns in the votes view. But I'll comment on that when I have my PO-hat on :-)

PR #316, opened 2018-05-03
Feature 156797118 Restrict Processing

roedesh:
- Added `restricted` option to Settings page
- New transaction form and like buttons are disabled for restricted users
- Names of restricted users are shown as `Hidden`
- Admin now only shows non restricted users
- Added decorators to handle transactions and users

screen shot 2018-05-03 at 12 20 48

screen shot 2018-05-03 at 12 20 31


app/controllers/users_controller.rb
```diff
@@ -26,7 +29,8 @@ def update
   end
 
   def autocomplete_search
-    @users = User.order(:name).where('lower(name) like ?', "#{params[:term]}%".downcase).where(deactivated_at: nil)
+    @users = User.order(:name).where('lower(name) like ?', "#{params[:term]}%".downcase)
```
ariejan: I know this was already in here, but I'd be better if we could move this query logic to the model. E.g. use something like `User.find_for_autocomplete(params[:term])`

app/dashboards/user_dashboard.rb
```diff
@@ -83,6 +83,10 @@ class UserDashboard < Administrate::BaseDashboard
   # Overwrite this method to customize how users are displayed
   # across all pages of the admin dashboard.
   def display_resource(user)
-    user.name
+    if user.restricted?
```
ariejan: `user.restricted? ? "Hidden" : user.name` is a common pattern for simple if statements.

db/schema.rb
```diff
@@ -143,12 +143,13 @@
     t.string   "uid"
     t.string   "avatar_url"
     t.string   "slack_name"
-    t.boolean  "admin",                  default: false
+    t.boolean  "admin"
```
ariejan: The "default: false" for admin has gone. Could this become an issue?

PR #317, opened 2018-05-04
Feature 156874214 Email Confirmation Status

roedesh:
Here is the PR for the changes made to the email confirmation status feature, shown yesterday during the demo.

Changes:
- Added a link to the warning message that will instantly resend the email, instead of redirecting you to the Devise form
- Made some changes to the _message.html partial and the close-message javascript function. Now flash messages can be closed individually
- Made soms changes to add_transaction spec to work with the new _message.html partial


config/routes.rb
```diff
@@ -14,6 +14,8 @@
   get :users, to: 'users#autocomplete_search', as: :users_autocomplete
 
   get :settings, to: 'users#edit', as: :user
+  get :resend_email_confirmation, to: 'users#resend_email_confirmation',
```
ariejan: This should not be a GET request IMHO.

http://thedailywtf.com/articles/WellIntentioned-Destruction

roedesh: I'll change it to a POST request

ariejan:
Or a PUT :-)

GET: Retrieve shit
POST: Create new shit
PUT: Update shit
DELETE: Throw away shit

roedesh:
Changed it. I chose POST since we CREATE some shit (a new confirmation token and a new email) :-)

PR #318, opened 2018-05-04
Changed votes to likes and made data more understandable

roedesh:
- Changed votes to likes
- Made data more understandable

screen shot 2018-05-04 at 13 52 42


app/controllers/users_controller.rb
```diff
@@ -1,21 +1,21 @@
 # frozen_string_literal: true
 class UsersController < ApplicationController
 
-  before_action :set_user, only: %i[edit update view_data view_transactions view_votes resend_email_confirmation]
+  before_action :set_user, only: %i[edit update view_data view_transactions view_likes resend_email_confirmation]
```
ariejan: At this point it might be more useful to use `except` instead of `only`.

PR #323, opened 2018-05-17
Bugfix wrong sender name

roedesh:
This PR will *most likely* fix the bug where the wrong sender name is used (we need to test it on staging, so Jeroen can access it). Previously `transaction_adder` would find the sender by using the `Api-Token` header that is provided in the request body. API v2 no longer needs this header, but it was still being used.

I've created another method which will find the user by using the `Authorization` header, which we are using right now. This makes sure that the right sender is added to the transaction.

### DIFF
--- a/app/controllers/api/v2/transactions_controller.rb
+++ b/app/controllers/api/v2/transactions_controller.rb
@@ -3,7 +3,7 @@ class Api::V2::TransactionsController < Api::V2::ApiController
   after_action :update_slack_transaction, only: [:like, :unlike]
 
   def create
-    @transaction = TransactionAdder.create_from_api_request(request.headers, params)
+    @transaction = TransactionAdder.create_from_v2_api_request(api_user, params)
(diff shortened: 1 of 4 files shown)


ariejan:
Any update on this?

roedesh:
I looked further into it. Because the Api-Header token is missing in the API V2 requests, the transaction_adder tries to look for a sender user where api_token is equal to a empty string, or in SQL:
`SELECT  "users".* FROM "users" WHERE "users"."api_token" IS NULL LIMIT 1`

So it fetches ALL users without a api_token and then the ordering (created_at) decides which user is returned. I guess that Vivian doesn't have a api_token on staging?

I created a spec for `transaction_adder.rb`, since the controller no longer uses the old method. I created specs for both `#create_from_api_request` and `#create_from_api_v2_request`.

I also wrote specs for `api/v2/transaction_controller_spec.rb`, which checks if the created transaction has the right sender name.

PR #324, opened 2018-05-25
Feature 156570222 Create Team

roedesh:
When a user logs in, three things can happen:

1. User has one team, gets redirected to that team page
2. User has multiple teams, choose one
3. User has no teams, show a message telling the user to either wait for a confirmation or create a team. Shows a button which links to a page where you can create a team.

Other changes
- Teams are now accessible by their own team urls. Kabisa for example: https://kudo-o-matic.nl/kabisa
- Users, transactions, balances and goals are scoped to their team. If you are currently working in Team A, you won't see anything from Team B.
- RSpecs updated to work with team functions 
- When a team is created, automatically create a balance, goals and a first example transaction.


app/controllers/api/v1/api_controller.rb
```diff
@@ -13,6 +13,10 @@ def api_token
     request.headers['Api-Token']
   end
 
+  def team_id
+    request.headers['Team']
```
ariejan: Why is this a request header and not a simple URL parameter?

app/controllers/api/v2/goals_controller.rb
```diff
@@ -1,10 +1,10 @@
 class Api::V2::GoalsController < Api::V2::ApiController
   def next
-    redirect_to api_v2_goal_path(Goal.next)
+    redirect_to api_v2_goal_path(Goal.next(@current_team_id))
```
ariejan: A common pattern in having things like a `current_team` and `current_user` is this:

```
def current_team
  @current_team ||= team_from_headers || team_from_some_other_place
end
```

This uses the `@current_team` if it has already been set this request; otherwise it will execute the other methods in order to find a team.

This can also be used in `ApplicationController` and made a helper (`helper_method :current_team`) which means you can use `current_user` and `current_team` throughout all views.

ariejan: The benefit of the above is this: you can use `current_team` everywhere without relying on a `before_action` or other mechanism to pre-load the team. Also, if this request does not _need_ the team at all, you save a precious call to the database.

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+class SessionsController < Devise::SessionsController
+  after_action :after_login, only: :create
+
+  def after_login
+    # TODO: if user is member of multiple teams, redirect to choice page (user story #156612953)
```
ariejan: Excellent TODO reminder!

app/controllers/slack_controller.rb
```diff
@@ -1,6 +1,8 @@
 class SlackController < ApplicationController
   include Slack::Messages
 
+  before_action :set_team, only: %i[action command reaction]
```
ariejan: So this can be removed (and you cannot forget it elsewhere) if you use the `current_team` method approach.

app/controllers/slack_controller.rb
```diff
@@ -12,12 +14,12 @@ def action
     transaction = Transaction.find(payload['callback_id'])
     user = User.find_by_slack_id(payload['user']['id'])
 
-    if transaction.present? && user.present?
+    if transaction.present? && user.present? && @current_team.member?(user)
```
ariejan: This is not all needed like this. Mainly because it's hard to read. 

L14: `Transaction.find` will throw an `ActiveRecord::RecordNotFound` if it cannot find the transaction. So the check in this if-statement is not needed. What _is_ needed may be proper error handling - but maybe that's taken care of elsewhere already?

L15: This, contrary to L14, will return `nil`.

Now, the magic happens with Ruby's Safe Navigation Operator. It allows you to call a method on an object that _might_ be `nil`. instead of a error, `nil` will be returned. so, L17 can be rewritten as:

`if user&.member_of?(current_team)`

This will check both if the user exists, and if it does, if it belongs to the current team. 

Also, it's way more readable :-)

app/controllers/teams_controller.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+class TeamsController < ApplicationController
```
ariejan: I would split this controller in two parts: creating/managing a team as a user and selecting which team to use. These are two separate concerns.

app/controllers/teams_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+  before_action :set_user, only: [:index, :create]
+
+  def index
+    # If user has only one team, redirect to that team's dashboard
+    if @user.teams.length == 1
```
ariejan: Two comments on this:

`@user.teams.length` will first query the database for relevant teams and instantiate `Team` activerecord objects for each one, and put them in an Array. `length` will return the length of that array. This is wasteful and a potential issue for larger numbers of users / teams.

Instead, you can use `#count`. When used on an ActiveRecord relation, it will execute a `COUNT` query and return an int. `@user.teams.count`

To go even further, Rails offers `#one?`, which will do the count query and then check if the result size is 1. 

`@user.teams.one?`

app/controllers/teams_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+      redirect_to dashboard_path(tenant: @user.teams[0].slug)
```
ariejan: I think `tenant` should be renamed through-out the app to `team` for consistency.

Redirecting to a team dashboard should look like:

`redirect_to team_dashboard_page(@team)`

`@team` could be `@user.teams.first`

app/controllers/teams_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+    @teams = @user.teams
+  end
+
+  def new
+    @team = Team.new(team_params)
+  end
+
+  def create
```
ariejan: This action does a lot of different things. 

Could you ask yourself: What object/class can I ask to setup a new team for this user? 

The answer is not this controller :-)

app/controllers/transactions_controller.rb
```diff
@@ -26,10 +28,10 @@ def show
   end
 
   def create
-    @transaction = TransactionAdder.create(params[:transaction], current_user)
+    @transaction = TransactionAdder.create(params[:transaction], current_user, @current_team)
```
ariejan: `TransactionAdder` is a helper class and adding more parameters here to support a use case seems dirty. 

Again: who could this controller ask to create a transaction for a team? How would that method signature look like?

app/controllers/transactions_controller.rb
```diff
@@ -39,20 +41,20 @@ def upvote
...
 
     respond_to do |format|
-      format.html {redirect_to :back}
+      format.html { redirect_to :back }
       format.js
     end
 
-    if Balance.current.amount >= Goal.next.amount
-      GoalReacher.check!
+    if Balance.current(@current_team).amount >= Goal.next(@current_team).amount
```
ariejan: I feel this should go in `TransactionAdder` or something? What do you think @roedesh?

app/jobs/export/create_export_job.rb
```diff
@@ -0,0 +1,94 @@
+# frozen_string_literal: true
+
+require 'rubygems'
+require 'zip'
+require 'open-uri'
+
+Export::CreateExportJob = Struct.new(:user, :team, :dataformat) do
+  def perform
```
ariejan: This is a job. The job of a job is to take arguments, load the appropriate data and start a process. This means that the job itself should not contain any logic - at all. 

Basically, you'd need to only supply an `export_id`:

```
def perform(export_id)
  export = Export.find(export_id)
  export.generate_zip!
end
```

ariejan: Next question is, is `Export` the right place to ask to create a ZIP file of data... or maybe we need some place else?

app/models/goal.rb
```diff
@@ -3,15 +3,20 @@ class Goal < ActiveRecord::Base
 
   belongs_to :balance
 
-  scope :goals, -> { where(balance: Balance.current).where(achieved_on: nil).order("amount ASC") }
-  scope :achieved, -> { where(balance: Balance.current).where.not(achieved_on: nil).order("amount ASC") }
+  def self.goals(team)
```
ariejan: Why supply team as a parameter?? 

`current_team.goals` should be the way to go.

ariejan: ActiveRecord will know how to handle that properly automagically.

app/controllers/teams_controller.rb
roedesh: A team_adder, like the transaction_adder?

app/controllers/transactions_controller.rb
roedesh: But isn't this required since Balances now belong to a team?

app/jobs/export/create_export_job.rb
roedesh: Should I create a new model, for example export_generator.rb, similar to the transaction_adder?

app/controllers/api/v1/api_controller.rb
ariejan: :+1:

app/controllers/teams_controller.rb
ariejan: Exactly. These "Service Objects" handle a specific use case, in this case creating and setting up a new team.

app/controllers/teams_controller.rb
roedesh: I made a team_selector_controller for choosing a team, all remaining actions stay in team_controller

app/controllers/teams_controller.rb
roedesh: Teams now get created by team_adder.rb

PR #328, opened 2018-05-31
Bug 157993814 api transactions not scoped

roedesh:
This PR fixes a number of issues:
- API V2
    - Transaction POST now adds the transaction to the right team
    - Transaction GET now only returns the transactions for a given team
    - Statisctics GET now only shows the statistics for a given team
- Added a spec for transaction_controller to check if the created transaction has the right Team ID
- Modifed specs of transaction and statistics to work with Team ID's

### DIFF
--- a/app/controllers/api/v2/statistics_controller.rb
+++ b/app/controllers/api/v2/statistics_controller.rb
@@ -10,11 +10,11 @@ class Api::V2::StatisticsController < Api::V2::ApiController
 
     @transactions_week = transactions_week.count
     @transactions_month = transactions_month.count
-    @transactions_total = Transaction.count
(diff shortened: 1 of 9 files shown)


app/controllers/api/v2/statistics_controller.rb
```diff
@@ -10,11 +10,11 @@ def general
 
     @transactions_week = transactions_week.count
     @transactions_month = transactions_month.count
-    @transactions_total = Transaction.count
+    @transactions_total = Transaction.where(team_id: current_team.id).count
```
ariejan: This is where my `current_team` suggestion comes in. If you write `current_team.transactions.count` there's not option to forget the `where` clause.

PR #329, opened 2018-05-31
Feature 156570222 create team changes

roedesh:
- Autocomplete now only returns team members
- Added team name to header on team pages

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -60,7 +60,7 @@ class UsersController < ApplicationController
   end
 
   def autocomplete_search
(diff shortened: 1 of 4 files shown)


app/controllers/users_controller.rb
```diff
@@ -60,7 +60,7 @@ def resend_email_confirmation
   end
 
   def autocomplete_search
-    @users = User.find_by_term(params[:term])
+    @users = current_team.users.find_by_term(params[:term])
```
ariejan: It might be useful to add a test for this (even better: write a test beforehand to illustrate the incorrect behaviour).

PR #331, opened 2018-06-06
Feature 157283177 Invite Existing User

roedesh:
- Added a manage page with invite form at `<team>/manage`
- Users can be invited by entering email addresses in a comma separated list
- Added pending invites to the team_selector, the user can accept or decline these invites. When a user accepts the invite, a new `TeamMember` will be created.
- Modified `teams/me` API endpoint to also include invites
- Added API endpoint for accepting or declining a invite
- Added controller specs for team_selector and team_invite
- Added feature specs for inviting users and accepting or declining invites


app/controllers/application_controller.rb
```diff
@@ -16,7 +16,13 @@ def check_team_membership
   end
 
   def current_team
-    @current_team ||= request.headers['Team'] || Team.find_by_slug!(params[:team])
+    unless @current_team
```
ariejan: Why did you rewrite this? To check if `params[:team]` exists? It might be better to use the old, more readable syntax and create methods for finding the team by param and header.

app/controllers/team_invite_controller.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+class TeamInviteController < ApplicationController
+  def create
+    @invites = TeamInviteSubmission.new(team_invite_params)
```
ariejan: Did you mean `@invite`?

app/models/team_invite.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class TeamInvite < ActiveRecord::Base
+  belongs_to :team
+  belongs_to :user
+
+  def complete?
+    !accepted_at.nil? || !declined_at.nil?
```
ariejan: `accepted_at || declined_at` will also do the job.

app/models/team_invite.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def accept
+    update_attribute(:accepted_at, Time.now)
```
ariejan: Be sure to wrap this in a transaction!

app/models/team_invite_submission.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  attr_accessor :emails
+
+  validates :emails, presence: true
+  validate :validate_emails
+
+  private
+
+  def validate_emails
+    email_list = emails.gsub(/\s+/, '').split(',')
```
ariejan: This looks oddly similar to the code in `TeamInviteAdder`. What's the difference? Is there a different use case for this class and `TeamInviteAdder`? I cannot see it. 

Also, the list is now split by `,`. This has several problems. Some people use spaces. Others copy paste from Outlook, which will use `;`. 

Also, the emails from the split list might still contain leading/trailing whitespace.

Also, what happens if a list is pasted with names included: `Ariejan de Vroom <ariejan@devroom.io>`? This is also common when copying from email clients.

app/models/user.rb
```diff
@@ -78,7 +79,16 @@ def self.find_by_term(term)
...
+    memberships.find_by_team_id(team.id).present?
+  end
+
+  def admin_of?(team)
+    membership = memberships.find_by_team_id(team.id)
+    membership&.admin?
+  end
+
+  def invited_to?(team)
+    TeamInvite.find_by_user_id_and_team_id(id, team.id).present?
```
ariejan: I'm not a big fan of magic helper methods. Why not:

`TeamInvite.where(user_id: id, team_id: team.id).any?` (again, this should run a COUNT query instead of fetching results). 

Also, there's already a relation from `User` to `team_invites` on L18. 

`team_invites.where(team_id: team.id).any?`

app/views/layouts/_header.html.haml
```diff
@@ -1,9 +1,14 @@
 .header-bg
   .header
     .img-container
-      %a{href: root_url}
-        = image_tag('kudo-o-matic-white.png', class:'header-img')
-        = image_tag('lizard.png', class:'mobile-img')
+      - if !current_team
```
ariejan: Try to avoid `!`, especially if you need to handle both cases. This reads way easier:

```
if current_team
 # this
else
 # that
end
```

app/views/team_selector/index.html.haml
```diff
@@ -8,6 +8,16 @@
...
           <i class="fa fa-sign-out" aria-hidden="true"></i> Logout
 
+    - flash.each do |key, value|
+      .message-container.show-message
+        .margin-bottom{:class => "#{key}-message"}
+          %span.message
+            .fa.fa-exclamation-circle
+            .message{:class => key}= value
+
+    - if @invites.size > 0
```
ariejan: `@invites.any?` or at least `@invites.count` ;-)

app/views/transactions/_profile-menu.html.haml
```diff
@@ -20,6 +20,11 @@
         %span.flex-item
           New team
         %i.fa.fa-plus
+      - if current_user.admin_of?(current_team)
```
ariejan: There are lots of call like this. That's fine, but the implementation of `admin_of` will keep doing queries (unless cached). Maybe we can help a bit and cache (memoize is the magic word) the function result.

app/controllers/application_controller.rb
roedesh: The reason I changed it is because in `layouts/_header.html` I check for the existence of `current_team`. If a team exists, show the team name, if not, show the Kudo-o-Matic logo.

The problem is, not every page has a `team` param, so for example `root_path` will throw an error because of `Team.find_by_slug!(params[:team])`, since there is no `team` param.

At first I wanted to solve this by just using `@current_team`, but in some cases it will be nill (for example after a invalid form submission where `@current_team` is not reset.

app/models/team_invite_submission.rb
roedesh: The reason I created TeamInviteSubmission is to make validating the invite form easier. When the user submits the form, it get's run trough TeamInviteSubmission validation, and after that it get's passed to the TeamInviteAdder, which will persist the data.

If there are errors, the ActiveModel validators will return the form errors. If there is an easier way, I'd like to know. I'm not so sure how TeamInviteAdder would be able to return form errors.

I haven't thought about the other email list formats. I'll look further into it.

app/controllers/team_invite_controller.rb
roedesh: I used @invites because it can contain multiple email addresses. Maybe I should change it to @email_addresses?

app/views/transactions/_profile-menu.html.haml
roedesh: I memoized `admin_of?`. I'm using [this method](https://www.justinweiss.com/articles/4-simple-memoization-patterns-in-ruby-and-one-gem#and-what-about-parameters).

I have added a `TODO` for when we implement the functionality to give or revoke admin rights to team members, because we will need to invalidate the cache when the admin value changes.

app/controllers/application_controller.rb
ariejan: Yes, I understand why you changed it. But your code would become more readable by combining the ternary if with helper methods that either return `nil` or a team.

app/controllers/team_invite_controller.rb
ariejan: Meh. If anything, change it to `@team_invite_submissions`.

app/controllers/team_selector_controller.rb
```diff
@@ -4,10 +4,11 @@ class TeamSelectorController < ApplicationController
 
   def index
     # If user has only one team, redirect to that team's dashboard
-    if current_user.teams.one?
+    if current_user.teams.one?  && current_user.team_invites.empty?
```
ariejan: Personally, I'd abstract this to a helper method: `show_current_user_team_selection?` But this is fine too.

app/controllers/team_selector_controller.rb
```diff
@@ -4,10 +4,11 @@ class TeamSelectorController < ApplicationController
...
       redirect_to dashboard_path(team: current_user.teams.first.slug)
     end
     @teams = current_user.teams
+    @invites = current_user.team_invites.where(accepted_at: nil).where(declined_at: nil)
```
ariejan: You use two `where` clauses here and I need to think to determine what invites are fetched here.

`current_user.team_invites.open` would be more revealing.

app/models/team_invite_submission.rb
ariejan: Ah, that makes sense. Instead of `Submission` call it `TeamInviteForm`, which reveals its purpose a bit better.

app/views/transactions/_profile-menu.html.haml
ariejan: Ah, that's a bit over the top :-) 

You could to it like this:

```
def admin_of?(team)
  @is_admin ||= #query here
end
```

For the lifetime of the object, the query result will be cached. Note that the object is discarded after each request, so for the next request we check the value again anyway. 

The reason this works is because repeated calls from views all need the same value for this request anyway.

PR #336, opened 2018-06-15
Feature 156569880 update team logo

roedesh:
- Added proper form error handling to create form
- Added update form
- Fixed a bug where the logo would be updated, even if there are logo validation errors.
- Show the logo of `current_team` in the header if there is one, else show the name.
- Added specs for creating and updating teams
- Added spec to show the logo bug has been fixed
- Added a navigation link in the profile menu to switch teams. This will only be visible if the user has at least 2 teams, or at least 1 team and 1 invite.


app/controllers/api/v2/team_invites_controller.rb
```diff
@@ -4,7 +4,7 @@ class Api::V2::TeamInvitesController < Api::V2::ApiController
   def update
     invite_id = params[:id]
     accept = params['accept']
-    @invite = TeamInvite.where(id: invite_id).where(accepted_at: nil).where(declined_at: nil).first
+    @invite = api_user.team_invites.open.where(id: invite_id).first
```
seamyd: looking for an id is cleaner with find(<id>) instead of where(....).first

app/controllers/api/v2/team_invites_controller.rb
```diff
@@ -4,22 +4,20 @@ class Api::V2::TeamInvitesController < Api::V2::ApiController
...
-      render json: { 'errors': [
-        {
-          title: 'Expired invite',
-          detail: 'This invite does not exist or has expired'
-        }
-      ] }
+      @invite.decline
+      render 'api/v2/teams/declined', status: :ok
     end
+  rescue ActiveRecord::RecordNotFound
```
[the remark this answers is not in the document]
roedesh: Is this better?

app/models/team_adder.rb
```diff
@@ -2,25 +2,27 @@
 
 class TeamAdder
```
thnukid: This file is in `app/models/` - I would expect that this inherits from `ActiveRecord::Base` - Why does the class not do that? 

The `create` and `save` hints that you re-creating rails built-in stuff.

app/views/devise/registrations/edit.html.haml
```diff
@@ -14,7 +14,7 @@
           #notice= notice
     %h2.settings-subtitle
       Details
-    = simple_form_for(resource, as: resource_name, url: registration_path(resource_name), html: { method: :put, class: 'form' }) do |f|
+    = form_for(resource, as: resource_name, url: registration_path(resource_name), html: { method: :put, class: 'simple_form form' }) do |f|
```
thnukid: What is the problem with using `simple_form_for` ?

app/views/teams/_general_form.html.haml
```diff
@@ -0,0 +1,22 @@
+%h2.settings-subtitle
+  General info
+- if @team.errors.any?
+  - @team.errors.full_messages.each do |msg|
+    .message-container.show-message
+      .margin-bottom{:class => "error-message"}
```
thnukid: ` .margin-bottom{:class => "error-message"}` i see no need for the `class:`

instead that should work `.margin-bottom.error-message`

app/models/team_adder.rb
roedesh: This object was created to remove the logic for creating a team, adding a member to that team and adding the first transaction away from the controller. TeamAdder does not reflect a database table. It is a "service object", as Ariejan calls them. It follows the same pattern as TransactionAdder.

app/views/teams/_general_form.html.haml
```diff
@@ -0,0 +1,22 @@
...
+        %span.message
+          .fa.fa-exclamation-circle
+          .message{class: :error}= msg
```
thnukid: `.message{class: :error}= msg` -> `.message.error = msg`

app/views/teams/_general_form.html.haml
```diff
@@ -0,0 +1,22 @@
...
+= simple_form_for @team, url: team_update_path(@team.slug), html: {class: 'form'} do |f|
```
thnukid: ` .form-inputs  .form-field = f.label :name` (line 11 till 20) all that markup will be rendered when you using `simple_form_for` unless the is a special reason, why you adding the `f.label` and `f.text_field`.

roedesh: You're right about the labels, my bad! The reason I add the fields is that I need to change the name attribute of the fields.

Usually the name will be something like `team[name]`, but because we have a URL parameter called `team`, it will cause a name conflict and throw an error.

app/controllers/api/v2/team_invites_controller.rb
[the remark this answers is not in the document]
thnukid: @roedesh Sorry I have edited my comment, the formatting got lost in the previous comment version

PR #387, opened 2018-08-28
Fixing delayed_job 

egonm12:
bin/delayed_job had to start manually. Because of this the transactions were not send to Slack until it was started.


ariejan:
Okay. But keep in mind DelayedJob is on the 'to delete' list in favour of Sidekiq, probably.

PR #391, opened 2018-08-30
Edit + delete a transaction

StefanPahlplatz:


app/controllers/transactions_controller.rb
```diff
@@ -37,11 +38,27 @@ def create
     end
   end
 
+  def destroy
+    @transaction = Transaction.find(params[:id])
```
ariejan: Why not use `set_transaction`?

app/controllers/transactions_controller.rb
```diff
@@ -37,11 +38,25 @@ def create
     end
   end
 
+  def destroy
+    if @transaction.destroy
+      flash[:success] = 'Succesfully removed transaction!'
```
egonm12: I would change this to delete

> Remove and Delete are defined quite similarly, but the main difference between them is that delete means erase (i.e. rendered nonexistent or nonrecoverable), while remove denotes take away and set aside (but kept in existence).

PR #398, opened 2018-08-31
Improved validation for removing balances

StefanPahlplatz:
Previously you were able to remove the only balance linked to a team as long as a current balance existsed, regardless of to which team is was linked. This is fixed with this pr.

### DIFF
--- a/app/models/balance.rb
+++ b/app/models/balance.rb
@@ -48,8 +48,10 @@ class Balance < ActiveRecord::Base
   private
 
   def ensure_a_current_balance_remains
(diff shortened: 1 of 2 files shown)


app/models/balance.rb
```diff
@@ -48,8 +48,10 @@ def self.time_left
   private
 
   def ensure_a_current_balance_remains
-    if Balance.where(current: true).count < 1
-      raise "Last current balance can't be removed from the system"
+    Team.all.each do |team|
```
ariejan: `team.balances`

PR #408, opened 2018-09-04
Created option to delete image from existing transaction

egonm12:


config/routes.rb
```diff
@@ -205,6 +205,8 @@
     post 'like/:id', to: 'transactions#upvote', as: :like
     post 'unlike/:id', to: 'transactions#downvote', as: :unlike
 
+    get 'transactions/:id/remove_image', to: 'transactions#remove_image', as: 'remove_transactions_image'
```
StefanPahlplatz: Is this still used?

PR #415, opened 2018-09-12
Removed slack from settings

StefanPahlplatz:


app/controllers/transactions_controller.rb
```diff
@@ -95,9 +95,11 @@ def update_slack_transaction
   end
 
   def check_slack_connection
-    if SLACK_IS_CONFIGURED && current_team.membership_of(current_user).slack_id.blank?
-      url = "<a href='#{settings_url(team: current_team.slug)}'>Connect your ₭udo-o-Matic account to Slack</a>"
-      flash.now[:warning] = url.html_safe
+    if current_team.name == "Kabisa"
```
egonm12: I would change this to `current_team.slug == "kabisa"` since a new team can also have the name "Kabisa". Slug is a unique id so it will never be the same.

app/views/users/edit.html.haml
```diff
@@ -8,28 +8,29 @@
           #notice= notice
     %h1.settings-title
       Settings
-    %h2.settings-subtitle
-      Slack
-    - if @user.slack_id_for_team(current_team).blank?
+    - if current_team.name == "Kabisa"
```
egonm12: Same here.

PR #418, opened 2018-09-18
Feature profile menu and team manage

egonm12:


app/assets/stylesheets/components/_forms.scss
```diff
@@ -17,27 +27,20 @@
...
     display: flex;
     margin-bottom: 1rem;
     align-items: center;
-    justify-content: center;
     input[type="text"],
     input[type="email"],
     input[type="password"] {
-      background-color: #ffffff;
-      border-radius: 0.25rem;
+      //    background-color: #ffffff;
```
StefanPahlplatz: Commet

app/assets/stylesheets/components/_header.scss
```diff
@@ -1,32 +1,136 @@
...
-  padding: $padding;
+//.container-header {
+//  padding: $padding;
+//
+//  .title, .subtitle {
+//    font-family: Orbitron;
+//    text-align: center;
+//    vertical-align: middle;
+//    margin-top: $margin-secondary;
+//  }
```
StefanPahlplatz: More comments

app/controllers/teams_controller.rb
```diff
@@ -35,6 +34,7 @@ def update
       flash[:success] = 'Successfully updated team!'
       redirect_to manage_team_path(@team.slug)
     else
+      # flash[:error] = 'Something went wrong, please try again!'
```
StefanPahlplatz: More comments

app/models/balance.rb
```diff
@@ -11,15 +11,17 @@
 #
 
 class Balance < ActiveRecord::Base
-  after_update :ensure_a_current_balance_remains
-  after_destroy :ensure_a_current_balance_remains
+  # after_update :ensure_a_current_balance_remains
```
StefanPahlplatz: More comments

spec/features/add_transaction_spec.rb
```diff
@@ -31,7 +31,7 @@
     fill_in 'user_email', with: user_2.email
     fill_in 'user_password', with: 'testpass'
     click_button 'Log in'
-
+    # first('.team-logo-selection').click
```
StefanPahlplatz: More comments

app/controllers/admin/balances_controller.rb
```diff
@@ -19,22 +19,29 @@ class BalancesController < Admin::ApplicationController
     # for more information
 
     def update
-      begin
-        @balance.update(balance_params)
-        redirect_to admin_balance_path
-      rescue Exception => e
-        flash[:error] = e.message
+      if balance_params['current'] == "0" && @balance.current
```
ariejan: What does this mean? I'd recommend extracting this conditional to a well-named private method.

ariejan: Also, if this is to prevent the current balance from being updated to `0` kudos, I'd recommend adding a validation/before_update of some sort on the Balance model instead.

app/controllers/admin/balances_controller.rb
```diff
@@ -19,22 +19,29 @@ class BalancesController < Admin::ApplicationController
...
+        end
       end
     end
 
     def destroy
-      begin
-        @balance.destroy
+      if @balance.current
+        flash[:error] = "Active balance can't be deleted"
+        redirect_to admin_balances_path
```
ariejan: It seems we're always redirecting to the same place, so lets' put it once, after all the conditionals.

Also consider that redirect is not an immediate action. E.g. it's not a 'return'.

app/controllers/balances_controller.rb
```diff
@@ -1,12 +1,20 @@
 # frozen_string_literal: true
 
 class BalancesController < ApplicationController
+  before_action :require_authorization
+
+  def require_authorization
```
ariejan: It seems this will be a recurring pattern. Maybe we can stuff it in a module and include it by default in the apps controller? (and skip it for controllers/actions where it doesn't make sense).

app/controllers/balances_controller.rb
```diff
@@ -1,12 +1,20 @@
...
+    redirect_to dashboard_path(team: current_team) unless current_user.admin_of?(current_team)
+  end
+
   def index
     @balances = current_team.balances.order('created_at DESC')
+    @current_balance = @balances.where(current: true)
```
ariejan: This might be an array of balances without an explicit `first`

app/controllers/balances_controller.rb
```diff
@@ -1,12 +1,20 @@
...
   end
 
   def show
     @balance = current_team.balances.find(params[:id])
+    @balances = current_team.balances.order('created_at DESC')
```
ariejan: Swap these around, as in `index`? You can then get `@balance` from `@balances`. I'm not sure if there's a performance gain there or not, but it looks better :-)

app/controllers/balances_controller.rb
```diff
@@ -29,7 +38,10 @@ def edit
 
   def update
     @balance = current_team.balances.find(params[:id])
-    if @balance.update_attributes(balance_params)
+    if balance_params['current'] == "0" && @balance.current
```
ariejan: Whoop. This conditional again. Use that helper methods here again.

app/controllers/balances_controller.rb
```diff
@@ -29,7 +38,10 @@ def edit
...
+      flash[:error] = "Active balance can't be deactivated!"
```
ariejan: I think you can do `redirect_to url, error: 'message'` as well. That's a bit more concise and reads better.

app/controllers/team_invite_controller.rb
```diff
@@ -19,6 +26,17 @@ def create
     end
   end
 
+  def delete
+    invite = TeamInvite.find(params[:id])
+    if invite.destroy
+      flash[:success] = 'Succesfully deleted invitation!'
+    else
+      binding.pry
```
ariejan: Maybe remove the debugging? :-)

app/controllers/team_selector_controller.rb
```diff
@@ -4,7 +4,7 @@ class TeamSelectorController < ApplicationController
 
   def index
     # If user has only one team, and no invites, go to the page of that team
-    if current_user.teams.one? && current_user.open_invites.empty?
+    if current_user.teams.one?  && current_user.open_invites.empty?
```
ariejan: What's with the extra whitespace?

app/controllers/teams_controller.rb
```diff
@@ -35,6 +34,7 @@ def update
       flash[:success] = 'Successfully updated team!'
       redirect_to manage_team_path(@team.slug)
     else
+      flash[:error] = 'Something went wrong, please try again!'
```
ariejan: I think `flash[:error] = ..` does not work in combination with `render`. `flash.now` is what you need to use I think.

app/controllers/users_controller.rb
```diff
@@ -4,7 +4,10 @@ class UsersController < ApplicationController
   before_action :set_user, except: %i[autocomplete_search]
   before_action :check_team_membership, only: %i[autocomplete_search]
 
+  def index; end
+
   def edit; end
+  def privacy; end
 
   def view_data
     @transactions_count = Transaction.all_for_user(@user).count
```
ariejan: Not something for now, but we really need to take a look at this syntax use: `@user.transactions` would have read much easier.

app/controllers/users_controller.rb
```diff
@@ -41,9 +44,11 @@ def download_export
 
   def update
     if @user.update(user_params)
-      redirect_to root_url
+      redirect_back(fallback_location: dashboard_path(team: current_team.slug))
+      flash[:succes] = "Changes saved successfully"
     else
-      render action: 'edit'
+      render action: 'index'
+      flash[:error] = "Something went wrong, please try again"
```
ariejan: I don't thinks this works (or is even needed), as the form will show specific validations errors anyways, right.

app/helpers/team_selector_helper.rb
```diff
@@ -0,0 +1,17 @@
+module TeamSelectorHelper
+  def team_option(team)
+    if !team.nil?
```
ariejan: A suggestion to this if statement:

```
return manage_team_path(team: team.slug) if team.present?
root_path
```

app/helpers/team_selector_helper.rb
```diff
@@ -0,0 +1,17 @@
...
+      manage_team_path(team: team.slug)
+    else
+      root_path
+    end
+  end
+
+  def team_option_text(team)
+    if !team.nil?
```
ariejan: See above.

app/controllers/admin/balances_controller.rb
egonm12: @ariejan this checks if the checkbox for current is unchecked (unchecked == "0") and if that balance is the 'current' balance. 
I designed it like this so you can never unset a current balance to not current. The only way to unset the balance as current is by setting another balance (in the same team) to current.

app/controllers/balances_controller.rb
egonm12: @ariejan Because there can only be one current balance in a team.

app/controllers/users_controller.rb
egonm12: @ariejan I am not sure what you mean by this? What has this to do with `def privacy; end`?

app/controllers/teams_controller.rb
egonm12: On second note I deleted the flash error completely since I can also show the validation error messages when something went wrong.

app/helpers/team_selector_helper.rb
egonm12: Much cleaner indeed but I forgot to delete these helper methods. They are not in use anymore.

app/controllers/team_selector_controller.rb
egonm12: @ariejan https://esolangs.org/wiki/Whitespace

app/controllers/users_controller.rb
egonm12: Ok clear, I was focussed on the green part 🙄

PR #419, opened 2018-09-19
Added mailtrap config to actionmailer

egonm12:


egonm12:
No it runs in staging. Should have checked this before commiting. I will close this PR for now.

PR #435, opened 2018-10-09
Transaction edit/delete improvements

egonm12:
* Team admins can now edit and delete all transactions
* Added timer so a transaction owner can edit/delete a transaction for a specific amount of time after its created
* Added exception on timer for team admins
* When a transaction is edited you will see `(edited)` in the transaction


app/controllers/transactions_controller.rb
```diff
@@ -176,8 +176,13 @@ def check_restricted
...
   def danger_methods
     if current_user.id != @transaction.sender_id
-      flash[:error] = "You're not authorized to perform this action!"
-      redirect_to root_url
+      (flash[:error] = "You're not authorized to perform this action!") && (redirect_to root_url) unless current_user.admin_of?(current_team)
+    end
+  end
+
+  def edit_or_delete_max_time
+    if @transaction.created_at < 15.minutes.ago
```
StefanPahlplatz: Shouldn't this be a setting?

app/controllers/transactions_controller.rb
```diff
@@ -176,8 +176,13 @@ def check_restricted
 
   def danger_methods
     if current_user.id != @transaction.sender_id
-      flash[:error] = "You're not authorized to perform this action!"
-      redirect_to root_url
+      (flash[:error] = "You're not authorized to perform this action!") && (redirect_to root_url) unless current_user.admin_of?(current_team)
```
ariejan: This is quite unreadable. You could move the `unless` to the top as a guard method (e.g. `return true if current_user.admin_of?(current_team)` followed by what happens when the guard fails.

app/controllers/transactions_controller.rb
ariejan: I agree with @StefanPahlplatz at least move it to a constant in `config/initializers` or something.

app/views/transactions/_timeline.html.haml
```diff
@@ -20,19 +20,22 @@
...
-              .transaction-dropdown{'data-id': transaction.id}
-                %a.transaction-option{href: edit_transaction_path(id: transaction.id, team: current_team)}
-                  .transaction-icon
-                    %span.fa.fa-pencil
-                  Edit transaction
-                .transaction-option
-                  .transaction-icon
-                    %span.fa.fa-trash
-                  = link_to 'Delete transaction', transaction_path(transaction, team: current_team), method: :delete, data: { confirm: 'Are you sure you want to delete this transaction?' }
+            - if transaction.created_at > 15.minutes.ago
```
ariejan: Here again, use a constant for the 15 minutes. Now you have a hard-coded value in two places in your code. You will curse yourself when you need to changes this to 10 minutes in a few weeks 🔥

app/views/transactions/_timeline.html.haml
```diff
@@ -20,19 +20,22 @@
               = image_tag transaction.receiver_image, class:'receiver flex-item'
           .transaction-right-corner
             %span.post-time-stamp
+              - if transaction.updated_at != transaction.created_at
```
ariejan: What does it mean for a transaction to be edited? 

For now this might be okay, but I can imagine the definition of 'edited' to change in the future. Maybe extract this into a method like `Transaction#edit_by_user?`?

app/views/transactions/_timeline.html.haml
```diff
@@ -20,19 +20,22 @@
...
-                %a.transaction-option{href: edit_transaction_path(id: transaction.id, team: current_team)}
-                  .transaction-icon
-                    %span.fa.fa-pencil
-                  Edit transaction
-                .transaction-option
-                  .transaction-icon
-                    %span.fa.fa-trash
-                  = link_to 'Delete transaction', transaction_path(transaction, team: current_team), method: :delete, data: { confirm: 'Are you sure you want to delete this transaction?' }
+            - if transaction.created_at > 15.minutes.ago
+              - if transaction.sender_id == @user.id || TeamMember.find_by_user_id(current_user).admin?
```
ariejan: A lot of logic again, move this to a helper or Transaction itself: `Transaction#editable_by?(@user)`

spec/controllers/transactions_controller_spec.rb
```diff
@@ -89,6 +102,18 @@
       end
     end
 
+    context 'as team admin' do
```
ariejan: I see no spec for disabling editing after 15 minutes.

PR #437, opened 2018-10-10
Issue rss feed

egonm12:
Currently all transactions that are created end up in Kabisa's Slack channel. @ ASML they just created a team and their transactions are also posted to Kabisa's Slack channel. I don't think that this is something we would want.


app/controllers/feed_controller.rb
```diff
@@ -3,7 +3,7 @@ class FeedController < ApplicationController
 
   def index
     request.format = :atom
-    @transactions = Transaction.last(25)
+    @transactions = Transaction.where(team: Team.find_by_slug("kabisa")).last(25)
```
ariejan: Meh. 

This hard-codes the RSS feed to include stuff form a single team. 

Better to create /team_name/feed.xml and scope properly per team.

egonm12:
This is a temporary fix to make sure that only Kabisa's transaction will be displayed on Slack. I will look into an option to create the RSS feed per team in different Slack teams.

PR #444, opened 2018-10-11
RSS Feed per team

egonm12:
To do before merge:
- [x] Resolve feedback

To do after deploying to production:

- [ ] Create tokens for existing teams
- [ ] Update RSS feed URL for Kabisa's Slack


app/controllers/feed_controller.rb
```diff
@@ -3,7 +3,8 @@ class FeedController < ApplicationController
 
   def index
     request.format = :atom
-    @transactions = Transaction.where(team: Team.find_by_slug("kabisa")).last(25)
+    redirect_to root_url if Team.find_by_slug(params[:team]).nil?
```
ariejan: This is expected to return XML/Atom. Redirecting to the `root_url` is not really what people expect. Instead maybe return a HTTP 404 with some plain text error message instead?

app/views/feed/index.atom.builder
```diff
@@ -6,7 +6,7 @@ atom_feed do |feed|
   end
 
   @transactions.each do |transaction|
-    feed.entry(transaction, url: root_url, updated: transaction.created_at) do |entry|
+    feed.entry(transaction, url: root_url + transaction.team.slug, updated: transaction.created_at) do |entry|
```
ariejan: Shouldn't `url` contain a direct link to the transaction? We now have a transaction detail page. (The old code was never updated I guess).

config/routes.rb
```diff
@@ -14,7 +14,7 @@
   post :resend_email_confirmation, to: 'users#resend_email_confirmation',
        as: :users_resend_email_confirmation
 
-  get :feed, to: 'feed#index'
+  get '/feed/:team', to: 'feed#index'
```
ariejan: Like the rest of the routes, we should nest `/feed` under team instead of the other way around. E.g. `/kabisa/feed.xml`.

spec/integration/feed_spec.rb
```diff
@@ -2,6 +2,10 @@
 require "rss"
 
 describe "/feed", type: :request do
+  before do
```
ariejan: `let(:team) { create :team ... }`

Using instance variables in tests is discouraged. `let` also lazy loads (so it's only created when used), which speeds up tests that don't need a team. Use `let!` instead of `let` if you need to eager load the variable.

spec/integration/feed_spec.rb
```diff
@@ -11,7 +15,7 @@
   context "given many transactions" do
     before do
       balance = create :balance, :current
-      @transactions = create_list(:transaction, 26, balance: balance, created_at: Date.yesterday.to_time)
+      @transactions = create_list(:transaction, 26, balance: balance, created_at: Date.yesterday.to_time, team_id: @team.id)
```
ariejan: Not something to change right now, but there are much cleaner and readable ways of creating a team with transactions using `FactoryBot`

app/controllers/feed_controller.rb
```diff
@@ -3,8 +3,12 @@ class FeedController < ApplicationController
 
   def index
     request.format = :atom
-    @transactions = Transaction.where(team: Team.find_by_slug("kabisa")).last(25)
+    if Team.find_by_rss_token(params[:rss_token]).nil?
```
ariejan: This feels a bit wonky. If _any_ team exists with the token, we continue to pick the selected team. This means I can access ALL teams with my team's RSS token.

```ruby
@team = Team.where(slug: params[:team], rss_token: params[:rss_token]).first
render('layouts/404', status: 4040, format: :html) and return if @team.nil?
# ...
```

db/migrate/20181011145946_add_rss_token_to_team.rb
```diff
@@ -0,0 +1,5 @@
+class AddRssTokenToTeam < ActiveRecord::Migration[5.0]
+  def change
+    add_column :teams, :rss_token, :string
```
ariejan: Awesome. I _do_ think you also need to trigger some action (could be manually after deployment) to actually set rss_tokens - they are not magically filled by themselves for existing Teams.

spec/integration/feed_spec.rb
```diff
@@ -2,36 +2,42 @@
...
     it "overrides updated timestamps so updates dont (re)appear in the feed" do
-      @transactions.first.touch
+      transactions.first.touch
       entries = parse_feed
 
-      expect(entries.first.updated.content).to eq(@transactions.first.created_at)
+      expect(entries.first.updated.content).to eq(transactions.first.created_at)
+    end
+
+    it 'renders 404 with wrong slug' do
```
ariejan: I would also add a test that you cannot access another team with your own team's token.

app/controllers/feed_controller.rb
egonm12: Good suggestion, I forgot indeed that this allowed a user to see the feed from all team's if is in possession of a valid rss token

db/migrate/20181011145946_add_rss_token_to_team.rb
egonm12: Will do this manually

app/controllers/feed_controller.rb
```diff
@@ -3,11 +3,9 @@ class FeedController < ApplicationController
 
   def index
     request.format = :atom
-    if Team.find_by_rss_token(params[:rss_token]).nil?
-      render 'layouts/404', status: 404, formats: :html
-    end
+    @team = Team.where(slug: params[:team], rss_token: params[:rss_token]).first
+    render('layouts/404', status: 404, formats: :html) && return if @team.nil?
```
ariejan: Use `&&` for binary operations (e.g. conditionals), use `and` for flow control (do this _and_ that).

egonm12:
```suggestion
    render('layouts/404', status: 404, formats: :html) and return if @team.nil?
```

PR #447, opened 2018-10-31
Graphql/users

egonm12:


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,10 @@
 # frozen_string_literal: true
 
 class ApplicationController < ActionController::Base
-  protect_from_forgery with: :exception
-  around_action :catch_not_found
-  before_action :authenticate_user!
-  before_action :configure_permitted_parameters, if: :devise_controller?
+  protect_from_forgery with: :null_session
+  # around_action :catch_not_found
```
ariejan: If it ain't needed, remove it.

app/controllers/concerns/authenticate_user.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+require "auth_token"
+
+module AuthenticateUser
+  Error = Class.new(RuntimeError)
+  InvalidHeader = Class.new(Error)
+  InvalidUser = Class.new(Error)
+  TokenExpired = Class.new(Error)
```
ariejan: What is this sorcery? Seriously, it's not common to use `Class.new` in Ruby. This _will_ confuse people. 

```ruby
class TokenExpired < RuntimeError; end
```

app/controllers/concerns/authenticate_user.rb
```diff
@@ -0,0 +1,59 @@
...
+  included do
+    @current_user = nil
+  end
+
+  def current_user
+    @current_user
+  end
+
+  def authenticate!
+    if auth_header.blank? || auth_header == "null"
```
ariejan: Maybe you can use `auth_header` to return `nil` if the String value is `"null"`. That way you keep all that logic in that particular method and the rest of your code can do "just ruby" by asserting nil or empty strings.

app/controllers/concerns/authenticate_user.rb
```diff
@@ -0,0 +1,59 @@
...
+        raise InvalidHeader.new "missing token" if t.nil?
+      end
+    end
+
+    def auth_payload
+      @_auth_payload ||= AuthToken.new.verify(auth_token)
+    end
+
+    def auth_ok?
+      !! auth_payload.dig(:ok)
```
ariejan: Don't put whitespace between negate and object. This makes is clearer what's begin negated.

E.g `!!auth_payload.dig(:ok)`

app/controllers/concerns/authenticate_user.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    def auth_expired?
+      !! (auth_payload.dig(:error) == :token_expired)
```
ariejan: Why the double negation here? `==` will already return a bool.

app/controllers/feed_controller.rb
```diff
@@ -1,11 +1,11 @@
 # frozen_string_literal: true
 
 class FeedController < ApplicationController
-  skip_before_action :authenticate_user!
+  # skip_before_action :authenticate_user!
```
ariejan: rm

app/controllers/feed_controller.rb
```diff
@@ -1,11 +1,11 @@
...
 
   def index
     request.format = :atom
-    @transactions = Transaction.where(team: Team.find_by_slug("kabisa")).last(25)
+    @posts = Post.where(team: Team.find_by_slug("kabisa")).last(25)
```
ariejan: maybe we can add a `# TODO ...` here to indicate we need to fix the hard coded `"kabisa"` here.

ariejan: And create a user story for it on the backlog.

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -38,7 +38,7 @@ def slack
 
     flash[:notice] = "Successfully #{current_user.slack_id.blank? ? 'connected to Slack!' : 'updated your Slack display name!'}"
 
-    SlackService.instance.set_general_channel_id(team)
+    # SlackService.instance.set_general_channel_id(team)
```
ariejan: rm

app/models/post.rb
```diff
@@ -0,0 +1,288 @@
...
+
+  def self.last_of_team(team)
+    Post.where(team_id: team.id).last
+  end
+
+  def liked_by_user?(user)
+    votes.find_by_voter_id(user.id).present?
+  end
+  #
+  # EMOJIES = [
```
ariejan: wot?

app/models/team.rb
```diff
@@ -21,23 +21,23 @@
...
 
 
-class Team < ActiveRecord::Base
+class Team < ApplicationRecord
   extend FriendlyId
   friendly_id :slug_candidates, use: :slugged
   after_create :setup_team
-  after_validation :logo_reverted?
-  has_attached_file :logo, styles: { thumb: "320x120" }
+  # after_validation :logo_reverted?
```
ariejan: rm

app/resources/api/v1/transaction_resource.rb
```diff
@@ -21,12 +21,12 @@ def votes_count
   end
 
   def image_url_original
-    # return nil if the transaction has no original image
+    # return nil if the post has no original image
```
ariejan: you could remove the useless comments. The code speaks for itself.

app/services/slack_service.rb
```diff
@@ -1,115 +1,115 @@
...
-    user["profile"]["display_name"] unless user.nil?
-  end
-
-  def set_general_channel_id(team)
-    return unless SLACK_IS_CONFIGURED
-
-    team.update_attribute(:channel_id, ENV["SLACK_CHANNEL"])
-  end
-end
+# # frozen_string_literal: true
```
ariejan: rm

ariejan: Yes. rm. 

if we're not using it and are planning on doing a re-write in the near future, just remove it. We have git to find it again if we need it.

config/initializers/devise.rb
```diff
@@ -256,7 +256,7 @@
       image_aspect_ration: "square",
       image_size: 512
 
-  config.omniauth :slack, ENV["SLACK_CLIENT_ID"], ENV["SLACK_CLIENT_SECRET"], scope: "team:read,users:read,identify,bot,commands"
+  # config.omniauth :slack, ENV["SLACK_CLIENT_ID"], ENV["SLACK_CLIENT_SECRET"], scope: "team:read,users:read,identify,bot,commands"
```
ariejan: rm

db/migrate/20181030103959_add_activity_to_post.rb
```diff
@@ -0,0 +1,13 @@
+class AddActivityToPost < ActiveRecord::Migration[5.2]
+  def self.up
```
ariejan: :+1:

app/controllers/concerns/authenticate_user.rb
egonm12: I'm not sure if I understand correctly what the benefit of this is, can you explain it a bit more in detail?

ariejan: Here you need to know that `auth_header` can be "null" (a string) that indicates a `nil` value. What I'm proposing it to put that bit of logic in the `auth_header` method instead. Using `auth_header` then either returns a legitimate value or `nil`. Now you need to know a specific detail about what `auth_header` will return and write a conditional for it. 

```ruby
if auth_header.blank?
  # stuff
end
```

is much clearer and intuitive.

app/controllers/feed_controller.rb
egonm12: Almost forgot about this since you were on holiday the day after I resolved the feedback. This PR should fix this: https://github.com/kabisa/kudo-o-matic/pull/444

app/models/post.rb
egonm12: Deleted because they were used to convert text to smileys in the frontend, but since all views are deleted this is not in use anymore. Will be deleted in the next commit.

app/controllers/concerns/authenticate_user.rb
ariejan: Well, not if `auth_header` can return `"null"` (a string) which should be interpreted as `nil`. In that case, the `auth_header` method should handle that case can convert `"null"` to `nil`. _Then_ you can make your suggested change.

(It's true that you can use `"null"` as auth_header and fail authentication, but it's better to recognize `"null"` for what it is, a `nil`.

egonm12: Well on second thought I think checking for `"null"` is unnecessary. Since it returns `nil` if Authorization header is not defined. And if it's a fake token, for example `"Bearer faketoken" it will a  throw a `RuntimeError`.

PR #453, opened 2018-11-09
Feature/161477282 guideline type

egonm12:


app/graphql/types/guideline_type.rb
```diff
@@ -0,0 +1,13 @@
+module Types
+  GuidelineType = GraphQL::ObjectType.define do
+    name 'Guideline'
+
+    field :id, !types.ID, 'The ID of the guideline'
+    field :name, !types.String, 'The name of the guideline'
+    field :kudos, !types.Int, 'The amount of kudos that is suggested'
```
StefanPahlplatz: So on `Goal` and `Post` it's `amount` but here it's `kudos`?

egonm12: I know it's really confusing at the moment so we should rename these column names asap

PR #455, opened 2018-11-11
Added mutation for resetting a user's password

egonm12:


app/graphql/mutations/user_mutation.rb
```diff
@@ -54,5 +54,25 @@ module Mutations
         )
       end
     end
+
+    field :resetPassword, Types::UserType do
+      description "Create a new user"
```
StefanPahlplatz:
```suggestion
      description "Reset a user's password"
```

PR #462, opened 2018-11-15
Added TeamMember roles

egonm12:


app/services/team_adder.rb
```diff
@@ -17,7 +17,7 @@ def self.create(params, current_user)
 
     def self.save(team, current_user)
       if team.save!
-        team.add_member(current_user, true)
+        team.add_member(current_user, 'admin')
```
ariejan: Maybe you can use `:admin` instead? As that's how you've defined the enum.

db/migrate/20181115082120_change_team_member_admin_column_to_int.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeTeamMemberAdminColumnToInt < ActiveRecord::Migration[5.2]
+  def change
+    change_column :team_members, :admin, :integer, :using => 'case when admin then 2 else 0 end'
```
ariejan: I think Postgres offers a native enum type that might be a better fit than integers. 

Maybe [this article](https://dev.to/amplifr/postgres-enums-with-rails-4ld0) can help you out.

PR #463, opened 2018-11-19
Added TeamInvite type and teamInvite fields on team and user query types

egonm12:


egonm12:
Somehow the `accept` and `decline` methods in the TeamInvite model don't update the field from `nil` to Time. I tried with `update_attribute` and `touch` and both work fine in development, but when I call them in my mutation tests it fails.

spec/graphql/mutations/team_invite_mutations/accept_invite_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+  let!(:user) { create(:user, email: "my@email.com") }
+
+  context 'authenticated user' do
+    describe "accept invite" do
+      it 'accepts the invite and creates a team member' do
+        team_invite = create(:team_invite, email: "my@email.com", team: team)
+
+        args = { team_invite_id: team_invite.id }
+        ctx = { current_user: user }
+        query_result = subject.fields["acceptInvite"].resolve(nil, args, ctx)
```
egonm12: This should update the `accepted_at` field of `team_invite`

ariejan: Could you add a `team_invite.reload` here? My guess is that you have an object here, update the data elsewhere and expect your own instance of the object to be modified. Meaning that you are not changing the state of the instance of `TeamInvite` here, but you're updating the data elsewhere.

PR #464, opened 2018-11-20
Added newPassword mutation to set a new password

egonm12:
Follow up for forgotPassword mutation.

To do:
- [x] Fix mutation test

Current schema check:
```
Change   Code           Description
───────  ─────────────  ────────────────────────────────────
FAILURE  FIELD_REMOVED  `Mutation.resetPassword` was removed
NOTICE   FIELD_ADDED    `Mutation.forgotPassword` was added
NOTICE   FIELD_ADDED    `Mutation.newPassword` was added
```


app/graphql/mutations/user_mutation.rb
```diff
@@ -55,13 +55,13 @@ module Mutations
       end
     end
 
-    field :resetPassword, Types::UserType do
+    field :forgotPassword, Types::UserType do
       description "Reset a user's password"
```
StefanPahlplatz: Description should also be updated.

app/graphql/mutations/user_mutation.rb
```diff
@@ -70,8 +70,34 @@ module Mutations
...
-        user
+        OpenStruct.new(
+          token: user.send_reset_password_instructions,
+          user: user
+        )
+      end
+    end
+
+    field :newPassword, Types::UserType do
+      description "Create new password if user forgot old one"
```
StefanPahlplatz: This is the description of the mutation, not the `newPassword` field.

app/graphql/mutations/user_mutation.rb
egonm12:
```suggestion
      description "Forgot a user's password"
```

PR #467, opened 2018-11-23
Added GoalReacher and PostMailer to PostMutation

egonm12:


app/helpers/goal_helper.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module GoalHelper
+  def kudos_to_next_goal(team)
+    Goal.next(team).amount - team.active_kudos_meter.amount
```
StefanPahlplatz: What happens when the last goal is reached?

app/mailers/post_mailer.rb
```diff
@@ -1,26 +1,31 @@
...
-    if receiver.name == ENV["COMPANY_USER"]
-      User.where.not(email: "").where.not(email: post.sender.email).where(deactivated_at: nil).each do |user|
-        suppress(Exception) { post_email(user, post).deliver_later if user.post_received_mail }
+    receivers.each do |receiver|
+      if receiver.email.present? && receiver.post_received_mail && receiver.deactivated_at.nil?
+        suppress(Exception) { post_email(receiver, post).deliver_now }
       end
     end
+
+    # TODO: Fix this method
```
StefanPahlplatz: ?

app/helpers/goal_helper.rb
egonm12: In the post mail: `- if !Goal.next(@post.team).nil?`

app/mailers/post_mailer.rb
egonm12: Will get fixed when we can mention the 'company user'.

PR #468, opened 2018-11-26
Added mutation to create team invites

egonm12:


app/graphql/mutations/team_invite_mutation.rb
```diff
@@ -50,5 +50,42 @@ module Mutations
...
+      type Types::TeamInviteType
+
+      resolve ->(_obj, args, ctx) do
+        current_user = ctx[:current_user]
+        team = Team.find(args[:team_id])
+
+        if current_user.blank?
+          raise GraphQL::ExecutionError.new("Authentication required")
+        end
+
```
StefanPahlplatz: What if the team does not exist?

egonm12: ` return team_invite if team_invite.save`

It just doesn't save

PR #469, opened 2018-11-28
Feature/162237552 multiple invites

egonm12:
You can now add more than 1 email to the mutation createInvite so you can send invites to multiple emails at once.


app/models/team_invite.rb
```diff
@@ -40,15 +40,18 @@ def self.open
...
-  private
-
   def duplicate?
-    invite_count = TeamInvite.where(email: email, team: team).count
-    raise "There is already an invite send to #{email}" if invite_count > 1
+    invites = TeamInvite.where(email: email, team: team).where(declined_at: nil).order("sent_at asc")
+
+    if invites.count > 1
+      invites.first.destroy!
+    end
```
StefanPahlplatz: Can't you do something like TeamInvite.where(email: email, team: team).where(declined_at: nil) `.delete_all()`?
Also shouldn't you also include accepted in the query?

app/graphql/mutations/team_invite_mutation.rb
```diff
@@ -77,14 +78,28 @@ module Mutations
           raise GraphQL::ExecutionError.new("Failed to create invite: Permission denied.")
         end
 
-        team_invite = TeamInvite.new(
-          email: args[:email],
-          team: team
-        )
+        if emails.blank?
+          raise GraphQL::ExecutionError.new("Emails can't be blank.")
```
ariejan: Is this shown on the front-end? Maybe "You didn't supply any email addresses" is a more clear error message.

app/graphql/mutations/team_invite_mutation.rb
```diff
@@ -77,14 +78,28 @@ module Mutations
...
+
+          unless team_invite.valid?
+            raise GraphQL::ExecutionError.new(team_invite.errors.full_messages.join(', '))
+          end
+
+          team_invites << team_invite
+        end
 
-        return team_invite if team_invite.save
+        team_invites.each(&:save)
```
ariejan: What happens when one of the team invites cannot be saved? Is this stuff wrapped in a transaction?

app/models/team.rb
```diff
@@ -62,7 +62,10 @@ def add_member(user, role = 'member')
   end
 
   def remove_member(user)
-    TeamMember.find_by_user_id_and_team_id(user.id, id).delete
+    transaction do
+      TeamMember.find_by_user_id_and_team_id(user.id, id).delete
+      TeamInvite.find_by_email_and_team_id(user.email, id).delete unless nil?
```
ariejan: uless `nil?` what?

app/models/team_invite.rb
```diff
@@ -40,15 +40,27 @@ def self.open
...
-  private
-
   def duplicate?
-    invite_count = TeamInvite.where(email: email, team: team).count
-    raise "There is already an invite send to #{email}" if invite_count > 1
+    # method to make sure only one invite per email exists at a time
+    invites = TeamInvite.where(email: email, team: team)
 
-    send_invite
+    if invites.count > 1
```
ariejan: Are you sure this work as you expect? 

The condition on L47 tests if you have more than one invite. E.g. 2 or more. If that's the case, then the destroying of the first invites makes no sense.

I think this can be written more concisely with guards instead. 

```ruby
invite = TeamInvite.where(email: email, team: team).first # AFAIK this should return 0 or 1 invites

send_invite and return if invite.nil? # Step 0: If you have no invites, ship one.

# At this point of execution, we know there is 1 invite.
if invite.accepted?
  # Do things and return
elsif invite.declined?
  # Do other things and return
end
```

Also, this can be done _before_ saving the new invite to the database. It's a good practice to not burden the db with unnecessary queries.

app/models/team.rb
egonm12: Unless nil? because if a user creates a team there will never be an invite for that user. So if you want to remove that specific member at some point you will get an error because invite does not exist.

ariejan: What are you calling `nil?` on?

app/models/team_invite.rb
ariejan: How did you address this?

app/graphql/mutations/team_invite_mutation.rb
ariejan: See my previous question.

app/models/team.rb
ariejan: You didn't answer my question. Why was this conversation resolved?

ariejan:
Nevermind, Github was not showing your more recent commits.

PR #471, opened 2018-11-28
Feature/162053589 post to non existing user

egonm12:


app/models/user.rb
```diff
@@ -132,6 +131,18 @@ def active_for_authentication?
 
   private
 
+  def email_required?
```
ariejan: Where are these used? You removed the email presence validation. 

`validates :email, presence: { unless: :virtual_user }` would also just work - no need for these helpers.

spec/graphql/mutations/post_mutations/create_post_spec.rb
```diff
@@ -8,21 +8,54 @@
...
-        team_id: team.id,
-        kudos_meter_id: kudos_meter.id
-      }
-    }
     let(:ctx) { { current_user: users.first } }
 
-    it "creates a new post" do
+    it "creates a new post with existing users" do
+      args = {
+          message: Faker::Lorem.sentence(3),
```
ariejan: If you don't need a message, don't stuff it full of data - it just takes up time and space unnecessarily. A simple `"message"` would suffice, unless you are testing parsing said message or something.

egonm12:
```suggestion
          message: 'message',
```

spec/graphql/mutations/post_mutations/create_post_spec.rb
```diff
@@ -8,21 +8,54 @@
...
+
+      expect do
+        subject.fields["createPost"].resolve(nil, args, ctx)
+      end.to change { Post.count }.by(1)
+     .and change { ActionMailer::Base.deliveries.count }.by(1)
+    end
+
+    it "creates a new post with non-existing users" do
+      args = {
+          message: Faker::Lorem.sentence(3),
```
ariejan: Also change these?

spec/graphql/mutations/post_mutations/create_post_spec.rb
```diff
@@ -8,21 +8,54 @@
...
+      expect do
+        subject.fields["createPost"].resolve(nil, args, ctx)
+      end.to change { Post.count }.by(1)
+      .and change { ActionMailer::Base.deliveries.count }.by(0)
+      .and change { User.count }.by(1)
+    end
+
+    it "creates a new post with existing and non-existing users" do
+      args = {
+          message: Faker::Lorem.sentence(3),
```
ariejan: Also change these?

ariejan:
Email validation is not broken, but already handled by Devise.

spec/graphql/mutations/post_mutations/create_post_spec.rb
egonm12:
```suggestion
          message: 'message',
```

spec/graphql/mutations/post_mutations/create_post_spec.rb
egonm12:
```suggestion
          message: 'message',
```

PR #474, opened 2018-11-29
Update seed

StefanPahlplatz:


db/seeds.rb
```diff
@@ -14,6 +14,7 @@
 )
 team.add_member(admin, 'admin')
 
+users = []
 ['Kabisa', 'Ariejan', 'Egon', 'Stefan', 'Ralph', 'Marijn', 'Guido', 'Managed Services'].each do |name|
```
egonm12: MS should have a '-' in between because `managed services@example.com` returns false
```suggestion
['Kabisa', 'Ariejan', 'Egon', 'Stefan', 'Ralph', 'Marijn', 'Guido', 'Managed-Services'].each do |name|
```

PR #477, opened 2018-12-04
Feature/162170944 user avatar

egonm12:
Users with no avatar set will have a gravatar set as their default avatar. At the moment only old users will have a custom avatar since registration used to be with google. This PR will: 

- Make a gravatar based on the user's email if the user has no avatar set
- Add a field to the UserType in the GraphQL API so we will be able to request the user's avatar

This PR does not include:
- Mutation to change avatar url
- Mutation to upload own image as url


app/models/user.rb
```diff
@@ -66,7 +60,15 @@ def first_name
   end
 
   def picture_url
-    avatar_url.presence || "/no-picture-icon.jpg"
+    return avatar_url.presence unless avatar_url.nil?
+
+    if email.blank?
+      gravatar = Digest::MD5::hexdigest(id.to_s).downcase
```
rhynix: If you want to write this in a shorter way this is possible:

`gravatar = Digest::MD5::hex digest(email.presence || id.to_s).downcase`

but I think this is also perfectly fine. 🙂.

app/models/user.rb
```diff
@@ -66,7 +60,15 @@ def first_name
   end
 
   def picture_url
-    avatar_url.presence || "/no-picture-icon.jpg"
+    return avatar_url.presence unless avatar_url.nil?
```
rhynix: `avatar_url.presence` suggests that `avatar_url` can be an empty string, don't you want to use `avatar_url.present?` in that case?

app/models/user.rb
```diff
@@ -2,14 +2,8 @@
 
 class User < ApplicationRecord
   validates :name, presence: true
-  validates :avatar, file_content_type: {
```
ariejan: If you remove this, we will not be able to remove uploaded avatar images any longer. It would be better to keep this for now and remove it only when we have fully phased out custom avatars.

app/models/user.rb
ariejan: I don't get this line at all, I'd expect an `avatar_url` to be returned if the field is not `nil`.

app/models/user.rb
ariejan: Why would you use the ID for gravatar? That will not work. 

If what you want to achieve is a default image from gravatar if the email does not exist, you can just as easily use a random number or something. You probably don't expose internal data through the MD5 hash, but I would not use it anyway.

config/application.yml
```diff
@@ -1,5 +1,7 @@
 defaults: &defaults
   token_expiration_time: <%= 1.month %>
+  gravatar_style: 'retro'
```
ariejan: Excellent choice!

spec/models/user_spec.rb
```diff
@@ -22,11 +22,6 @@
       expect { user.destroy }.to change { PostReceiver.count }
     end
 
-    it "should have one attached avatar" do
```
ariejan: This still holds for current accounts with an uploaded avatar.

app/models/user.rb
egonm12: Well it was something I added when migrating to ActiveStorage but the fact is that there was never something like having an uploaded avatar. A user's avatar has always been an url. That's why there is this field called `avatar_url`.

spec/models/user_spec.rb
egonm12: See my comment on avatar validation

app/models/user.rb
egonm12: @rhynix Ok I will change this to `return avatar_url if avatar_url.present?`

@ariejan avatar_url is currently used for your Gmail profile image. It is a link like this: https://lh4.googleusercontent.com/-PuxF7-XmxLw/AAAAAAAAAAI/AAAAAAAAAKQ/S_fX_1DAcYE/s512/photo.jpg ).

app/models/user.rb
egonm12: @ariejan If you want that every (virtual) user without an email has the same default gravatar than I don't need this check, since `email` is by default an empty string which also results in a gravatar.

app/models/user.rb
ariejan: :+1: Excellent, remove it then

PR #478, opened 2018-12-04
Added delete post mutation

egonm12:
I'm currently working on a better way of authorization in a central place. This mutation should prevent users that are not the sender of the post from deleting the post. The next users are authorized to delete a post:
- The sender of the post (max 15 minutes)
- The team admin (no time limit)
- The app admin (no time limit)


app/graphql/mutations/post_mutation.rb
```diff
@@ -67,5 +67,33 @@ module Mutations
...
+
+        if current_user.blank?
+          raise GraphQL::ExecutionError.new("Authentication required")
+        end
+
+        post = Post.find(args[:id])
+
+        break unless post.sender == current_user || current_user.admin_of?(post.team) || current_user.admin?
+
+        break if post.created_at < Settings.deletable_time_for_post && post.sender == current_user
```
rhynix: This means that an admin can't delete his or her own post after 15 minutes, right? Do you want this?

Perhaps you can make this a bit simpler by extracting the logic that determines if the user can delete the post in a new method. E.g.:

```
def can_user_delete_post(user, post)
  return true if user.admin?
  return true if user.admin_of?(post.team)
  return post.created_at < Settings.deletable_time_for_post if post.sender == user

  false
end
```

don't know where you would place this though. If you have a lot of this kind of logic, it might make sense to look into something like [pundit](https://github.com/varvet/pundit).

app/graphql/mutations/post_mutation.rb
```diff
@@ -67,5 +67,33 @@ module Mutations
...
+
+        if post.destroy
+          return post.id
```
rhynix: You don't need an explicit return here.

app/graphql/mutations/post_mutation.rb
```diff
@@ -67,5 +67,33 @@ module Mutations
...
+        else
+          GraphQL::ExecutionError.new(post.errors.full_messages.join(', '))
```
rhynix: [On this line](https://github.com/kabisa/kudo-o-matic/pull/478/files#diff-b9dfc89fe45f11475a16c9b1c28148d2R82) you raise an error, here you return it. Is this on purpose?

app/graphql/mutations/post_mutation.rb
```diff
@@ -67,5 +67,33 @@ module Mutations
...
+
+        current_user = ctx[:current_user]
+
+        if current_user.blank?
+          raise GraphQL::ExecutionError.new("Authentication required")
+        end
+
+        post = Post.find(args[:id])
+
+        break unless post.sender == current_user || current_user.admin_of?(post.team) || current_user.admin?
```
ariejan: Shouldn't you raise a `GraphQL::ExecutionError` here as well, like on L82?

config/application.yml
```diff
@@ -1,8 +1,10 @@
 defaults: &defaults
   token_expiration_time: <%= 1.month %>
+  deletable_time_for_post: <%= 15.minutes.ago  %>
```
ariejan: This won't work. This file is evaluated when the rails server boots. Technically it will work - for 15 minutes. After that the functionality breaks. Set this to `15.minutes` instead and calculate the exact time when making the comparison.

app/graphql/mutations/post_mutation.rb
ariejan: 1. You already checked the `post.sender == current_user` on L87. 
2. See my comment on `config/application.yml`. `Settings.deletable_time_for_post` is evaluated at boot. Instead use something like `break if created_at < (Time.now - Settings.deletable_after)` and set `deletable_after` to something like `15.minutes`.

egonm12: I found out my test was corrupt (Ctrl+C, Ctrl+V). Good you noticed!

app/graphql/mutations/post_mutation.rb
egonm12: This should `raise` as well. 
```suggestion
          raise GraphQL::ExecutionError.new(post.errors.full_messages.join(', '))
```

config/application.yml
egonm12: 👍

app/graphql/mutations/post_mutation.rb
egonm12: Will do!

app/graphql/mutations/post_mutation.rb
egonm12: Fixed!

app/models/post.rb
```diff
@@ -31,6 +31,14 @@ class Post < ApplicationRecord
 
   delegate :name, to: :sender,   prefix: true
 
+  def self.editable_time
+    if Rails.env.development?
+      DateTime.now - 1.year
+    else
+      DateTime.now - 15.minutes
```
ariejan: Eek! Hard-coded settings. 

You can still pull the `15.minutes` from the app yml.

app/graphql/mutations/post_mutation.rb
ariejan: Did you add this change or not? Github shows me a suggested change only.

PR #487, opened 2018-12-21
Refactoring for GraphQL class-based syntax and add authorization

egonm12:
According to the roadmap of graphql-ruby `.define` is going to be removed in v2.0, which means we need to switch to class-style instead of .define-style type definitions. We are currently using graphql-ruby 1.8.11

See the upgrade notes here: http://graphql-ruby.org/schema/class_based_api

This release will:
- Add authorization for all queries and mutations
- Refactor all types, resolvers and mutations to the class-based syntax


ariejan:
Does this have any implications for the front-end?

egonm12:
> Does this have any implications for the front-end?

Yes, working on that as I speak. We are not merging this before everything is implemented on the front-end. https://github.com/kabisa/kudos-frontend/pull/59

PR #489, opened 2019-01-14
Refactor directory structure of mutations

egonm12:
FYI: Unfortunately there is no option to rename the mutations like `Post::Create`, `Team::Create`, `TeamInvite::Create` since this will result in a duplicate type definition of the mutation payload (e.g. 3 times `createPayload`. This is a known issue (https://github.com/rmosolgo/graphql-ruby/issues/1505)


app/graphql/mutations/post/create_post.rb
```diff
@@ -12,32 +12,29 @@ class CreatePostMutation < BaseMutation
...
     def resolve(**kwargs)
-      team = Team.find(kwargs[:team_id])
+      team = ::Team.find(kwargs[:team_id])
       receivers = []
       receiver_ids = kwargs[:receiver_ids]
 
       unless receiver_ids.blank?
         receiver_ids.each do |receiver|
-          receivers << User.find(receiver)
+          receivers << ::User.find(receiver)
```
ariejan: Why prefix this one with `::` and not the `User` on L29?

app/graphql/mutations/post/create_post.rb
```diff
@@ -12,32 +12,29 @@ class CreatePostMutation < BaseMutation
...
 
       null_receivers = kwargs[:null_receivers]
 
       unless null_receivers.blank?
         null_receivers.each do |receiver|
-          user = User.new(
-            name: receiver,
-            virtual_user: true
-          )
+          user = User.new(name: receiver, virtual_user: true)
```
egonm12:
```suggestion
          user = ::User.new(name: receiver, virtual_user: true)
```

app/graphql/mutations/post/create_post.rb
egonm12: @ariejan missed that one. Send this PR to you because of the information I wrote in the description.

app/graphql/types/mutation_type.rb
```diff
@@ -2,22 +2,32 @@
...
-    field :reset_password, mutation: Mutations::ResetPasswordMutation
-
-    field :create_post, mutation: Mutations::CreatePostMutation
-    field :delete_post, mutation: Mutations::DeletePostMutation
-
-    field :create_team, mutation: Mutations::CreateTeamMutation
-    field :create_team_invite, mutation: Mutations::CreateTeamInviteMutation
-
-    field :toggle_like_post, mutation: Mutations::ToggleLikePostMutation
+    ##############
```
ariejan: What's all this? 

`### Posts` and three empty lines above it would suffice.

app/graphql/mutations/post/create_post.rb
ariejan: Ah. 

Well, this is an okay refactoring, but if it's clearer to keep the old design, just do that.

PR #501, opened 2019-01-16
Added check for team member role on update and delete

egonm12:
When deleting a team member or updating a role there should be a default check that gives an error when that deleted or updated member is the last admin. Best option is to raise an error that notifies the user that there needs to be at least 1 other team admin.
This will not cause issues if we build a feature to delete a whole team.

**NOTE: Commit message is wrong**


ariejan:
Am I right that you have applied this guard at different locations in the code? IMHO there should be only one place where a role can change (`TeamMember`, right?) and then use Rails' life cycle callbacks to handle this guard. If you don't, it's very easy for other developers, later, to write a new features and _forget_ to put in the guard.

PR #509, opened 2019-03-26
Revert bundler back to stable version

egonm12:


egonm12:
Fixed by updating and rebuilding the dokku host base image.

